PR #63, opened 2014-12-23
Fixed isActual implemenation

glesage:
Per https://github.com/strongloop/loopback/issues/468

However, I did notice that this does not work with json object mapping (I noticed this was a problem before as well: https://github.com/strongloop/loopback-connector-mysql/pull/29)

So if you have a property in a model as such:
```
{
  "name": "MyModel",
  "base": "PersistedModel",
  "properties": {
    "myProperty": {
      "type": "number",
      "mysql":{ "dataType":"float" }
    }
  }
  ...
}
```

Then isActual will ALWAYS return that your database is out of sync /: But I think this has more to do with the object mapping fix than this function correct?

Also, I was unable to run the test on my local, it wouldn't fully take in my MySql credentials even after hard coding them into init.js ... not sure how to do so I'll let the automated testing/builder do this

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -691,10 +691,10 @@ MySQL.prototype.createTable = function (model, cb) {
  * @param {String[]} [models] A model name or an array of model names. If not present, apply to all models
  * @param {Function} [cb] The callback function
  */
-MySQL.prototype.isActual = function(cb) {
+MySQL.prototype.isActual = function(models, cb) {
   var self = this;
   var ok = false;
-  async.each(Object.keys(this._models), function(model, done) {
+  async.each(Object.keys(models), function(model, done) {
     var table = self.tableEscaped(model);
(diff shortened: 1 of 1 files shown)


lib/mysql.js
```diff
@@ -691,10 +691,10 @@ MySQL.prototype.createTable = function (model, cb) {
  * @param {String[]} [models] A model name or an array of model names. If not present, apply to all models
  * @param {Function} [cb] The callback function
  */
-MySQL.prototype.isActual = function(cb) {
+MySQL.prototype.isActual = function(models, cb) {
   var self = this;
   var ok = false;
-  async.each(Object.keys(this._models), function(model, done) {
```
raymondfeng: The models arg is optional. Can you use the same logic at https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/postgresql.js#L1009-L1020?

glesage:
Alright done

PR #66, opened 2015-01-08
fixed setting the value of a non-nullable, non-required field to NULL

mihailignatenko:
Also added check if property actually has `mysql.nullable` option

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -272,6 +272,11 @@ MySQL.prototype.toFields = function (model, data) {
       if (undefined === value) {
         return;
       }
+      // skip property if it cannot be NULL, but `toDatabase` function returns NULL
+      if(props[key].mysql && props[key].mysql.nullable &&
+        props[key].mysql.nullable.substring(0, 1) === 'N' && value === 'NULL') {
+        return;
+      }
       fields.push(self.columnEscaped(model, key) + ' = ' + value);
     }
   }.bind(this));


slnode:

raymondfeng:
@mihailignatenko Thanks for the patch. Would you please add a test case?

lib/mysql.js
```diff
@@ -272,6 +272,11 @@ MySQL.prototype.toFields = function (model, data) {
       if (undefined === value) {
         return;
       }
+      // skip property if it cannot be NULL, but `toDatabase` function returns NULL
+      if(props[key].mysql && props[key].mysql.nullable &&
```
raymondfeng: We should test `|| props[key].mysql.nullable === true` too.

raymondfeng:
@mihailignatenko Please also sign CLA - https://cla.strongloop.com/agreements/strongloop/loopback-connector-mysql

raymondfeng:
@mihailignatenko ping

lib/mysql.js
mihailignatenko: @raymondfeng do you mean that `nullable` can be set to `true` or `false`, not only `"Y"` or `"N"`

raymondfeng: I intend to normalize the `nullable` to be a boolean. That's why I suggest that you cover `true` too.

mihailignatenko:
#92

jannyHou:
@mihailignatenko Are you still working on this pr? Could you rebase the code and add some test cases? thanks.

slnode:

slnode:

slnode:

slnode:

jannyHou:
Thank you so much for contribution, unfortunately the function fixed in this PR is removed, and the fix seems out of date now. I am closing it, feel free to reopen if there is update, thanks for understanding.

PR #67, opened 2015-01-09
Use mysql.escape/escapeId()

raymondfeng:
/to @bajtos or @ritch

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -171,7 +171,7 @@ MySQL.prototype.query = function (sql, callback) {
     }
     if (self.settings.createDatabase) {
       // Call USE db ...
-      connection.query('USE `' + db + '`', function (err) {
+      connection.query('USE ' + client.escapeId(db), function (err) {
         if (err) {
           if (err && err.message.match(/(^|: )unknown database/i)) {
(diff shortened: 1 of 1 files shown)


lib/mysql.js
```diff
@@ -871,7 +873,8 @@ MySQL.prototype.alterTable = function (model, actualFields, actualIndexes, done,
   function actualize(propName, oldSettings) {
     var newSettings = m.properties[propName];
     if (newSettings && changed(newSettings, oldSettings)) {
-      sql.push('CHANGE COLUMN `' + propName + '` `' + propName + '` ' +
+      var pName = self.client.escapeId(propName);
+      sql.push('CHANGE COLUMN ' + pName + ' ' + pName + ' ' +
```
ritch: Was this change intentional?

raymondfeng: Yes. Why do you ask?

ritch:
The changes look good. I am not deeply familiar with this code so I'm not sure if you've missed any strings.

lib/mysql.js
ritch: Looks like the arg was propName. Just wanted to confirm.

PR #68, opened 2015-01-15
Changed default type mapping

glesage:
Fix for https://github.com/strongloop/loopback/pull/946

Instead of a manual change in Loopback: https://github.com/strongloop/loopback/pull/980


slnode:

PR #69, opened 2015-01-15
Fixed missing 'ok' declaration

glesage:
Not sure when that ok disappeared.. but I just added it back


slnode:

PR #71, opened 2015-01-28
Include tests of persistence hooks from juggler.

bajtos:
Requires https://github.com/strongloop/loopback-datasource-juggler/pull/403.

Connect strongloop/loopback-datasource-juggler#367

/to @raymondfeng please review

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "debug": "~2.0.0"
   },
   "devDependencies": {
-    "loopback-datasource-juggler": "1.x.x",
+    "loopback-datasource-juggler": "^2.15.0",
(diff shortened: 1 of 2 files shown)


bajtos:
BTW we should strongly consider describing the instructions for setting up a local MySQL instance for the tests. This pull requests adds another 3 minutes to the already long time needed for tests to complete on my machine.

bajtos:
I have reworked the test suite to use a single data-source instance for all tests, I was hoping that reusing the same MySQL connection would speed up the test. It did not :(

raymondfeng:
LGTM

PR #74, opened 2015-02-17
NULL date instead of NaN

rudzon:
An invalid date cause ```dateToMysql``` to generate ```NaN-NaN-NaN NaN:NaN:NaN``` as a result and you receive ```0000-00-00 00:00:00``` timestamp.

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -336,6 +336,9 @@ MySQL.prototype.toDatabase = function (prop, val, forCreate) {
     if (!val.toUTCString) {
       val = new Date(val);
     }
+    if (isNaN(val.getTime())) {
+      return 'NULL';
+    }
     return '"' + dateToMysql(val) + '"';
   }
   if (prop.type === Boolean) {


slnode:

raymondfeng:
I wonder if the issue has already been fixed by https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L484-L490. Can you add a test case?

rudzon:
Passing ```NaN``` date results in ```0000-00-00 00:00:00``` in db , which is actually ```NaN``` for https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L484-L490 .
When the ```NaN``` for type ```Number``` is implicitly casted to ```NULL```.

raymondfeng:
@rudzon I understand the effect of invalid Date object. What I'm trying to understand how the invalidate can be passed to the mysql connector since https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L484-L490 already throws.

rudzon:
@raymondfeng When using updateAll method over REST, the ```data``` is passed raw here https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1540. This ```try``` https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1530 is checking ```where``` but not the ```data```

raymondfeng:
Maybe we should enhance updateAll to coerce the data per property definitions.

raymondfeng:
Closing this one to favor https://github.com/strongloop/loopback-datasource-juggler/issues/470

PR #75, opened 2015-02-18
updateOrCreate should allow updating the instance property value with ar...

krv783:
...rays of nested objects

When updating a model instance, which has a property containing an array of nested Objects, and when using ```PUT /Modelname``` and including the instance ```id``` in the request payload, the value of the property containing the array of nested Objects should be updated with a string representation of the array of Objects. instead of updating the MySQL column with ```[object Object]```

### DIFF
--- a/test/mysql.test.js
+++ b/test/mysql.test.js
@@ -82,6 +82,38 @@ describe('mysql', function () {
 
     });
   });
+  
+  it('updateOrCreate should allow updating the instance property value with arrays of nested objects', function (done) {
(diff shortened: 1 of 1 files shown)


slnode:

raymondfeng:
ok to test

raymondfeng:
@krv783 Not sure why you closed it. Is it by mistake?

krv783:
I added the test code to the wrong mysql test file. You linked to the one
under the BLOB folder, which is not the one I initially edited.

PR #77, opened 2015-02-26
add $where to support comparison between fields

violet-day:
see [loopback-datasource-juggler#444](https://github.com/strongloop/loopback-datasource-juggler/issues/444)

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -305,7 +305,7 @@ MySQL.prototype.toDatabase = function (prop, val, forCreate) {
     var operator = Object.keys(val)[0]
     val = val[operator];
     if (operator === 'between') {
-      return  this.toDatabase(prop, val[0]) +
+      return this.toDatabase(prop, val[0]) +
         ' AND ' +
         this.toDatabase(prop, val[1]);
     } else if (operator === 'inq' || operator === 'nin') {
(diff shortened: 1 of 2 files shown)


slnode:

lib/mysql.js
```diff
@@ -465,72 +466,76 @@ MySQL.prototype._buildWhere = function (model, conds) {
...
-        case 'like':
-          sqlCond += ' LIKE ';
-          break;
-        case 'nlike':
-          sqlCond += ' NOT LIKE ';
-          break;
-      }
-      sqlCond += (condType === 'inq' || condType === 'nin') ? '(' + val + ')' : val;
-      cs.push(sqlCond);
+    if (key === '$where') {
```
raymondfeng: The $ prefix is not consistent with other operators.

lib/mysql.js
```diff
@@ -465,72 +466,76 @@ MySQL.prototype._buildWhere = function (model, conds) {
...
+      cs.push(conds[key]);
```
raymondfeng: This could be vulnerable to SQL injection. The other thing is that we allow a model property to be mapped a column that has different name.

lib/mysql.js
violet-day: The key is the field.I use `$where` to keep difference with normal field

lib/mysql.js
violet-day: I consider about this.At present,I have 2 way.
1.In `$where` to use `$` or `this` prefix.
```js
Model.find({
  filter:{
    $where:'this.star>this.watch'//or :$star>$watch
  }
})
```
2.Or we can change the value.
```js
Model.find({
  filter:{
    star:{
      gt:'this.watch'//or :$watch
    }
  }
})
```
Which way you prefer~?

jannyHou:
@violet-day Thank you for contribution. Could you sign CLA then rebase your code?

slnode:

slnode:

slnode:

slnode:

jannyHou:
Thank you so much for contribution, unfortunately this PR is out of date and without CLA signed. I am closing it, please feel free to reopen if there is an update, thanks for understanding.

PR #78, opened 2015-03-02
Allow models backed by MySQL to reference mongodb ObjectID

raymondfeng:
/to @bajtos or @ritch 

When related models, for example, Customer and Order, are backed by MongoDB and MySQL, the foreign key (customerId) on MySQL can be typed as ObjectID. The PR allows the column to be saved/loaded correctly.

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -323,6 +323,9 @@ MySQL.prototype.toDatabase = function (prop, val, forCreate) {
   if (!prop) {
     return this.client.escape(val);
   }
+  if (prop.type === String) {
+    return this.client.escape(String(val));
+  }
   if (prop.type === Number) {
     if (isNaN(val)) {
       val = null;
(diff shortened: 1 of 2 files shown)


bajtos:
I am confused. Why would you use MongoDB's ObjectId type when storing the data to MySQL? Is this about some sort of MongoDB-to-MySQL replication?

I am not confident that the single new test covers all changes you made in the production code.

lib/mysql.js
```diff
@@ -345,18 +348,23 @@ MySQL.prototype.toDatabase = function (prop, val, forCreate) {
...
+  return serializeObject(val);
 };
 
+function serializeObject(obj) {
+  var str = JSON.stringify(obj);
+  if (str && str[0] === '"' && str[str.length - 1] === '"') {
+    // Remove the " for string values
+    str = str.substring(1, str.length - 1);
+  }
+  return "'" + str + "'";
```
bajtos: This creates a SQL injection vulnerability because `JSON.stringify` is not escaping single quotes.

```
// io.js 1.4.2
> console.log(JSON.stringify("a'b"))
"a'b"
```

raymondfeng:
This is the case that a model backed by MySQL (`Order`) has a `belongsTo` relation to a model backed by MongoDB ( `Customer`). LoopBack derives the type of foreign key property (`Order.customerId`) from the type of the primary key (`Customer.id`), which is ObjectID.

raymondfeng:
I fixed the SQL injection concern.

bajtos:
I am not confident the change is free of unintended side effects, but I don't see any obvious problem either. FWIW, LGTM.

bajtos:
Also please check why the tests are failing on Jenkins CI before landing.

panduroab:
@raymondfeng @bajtos when will be released the new version of loopback-connector-mysql with this solution?

raymondfeng:
Yes, we're testing and preparing a release.

PR #79, opened 2015-03-02
Query string length for schema in characters in addition to bytes

partap:
Add a field for string length when querying model properties.

### DIFF
--- a/lib/discovery.js
+++ b/lib/discovery.js
@@ -144,8 +144,7 @@ function mixinDiscovery(MySQL) {
       sql = paginateSQL('SELECT table_schema AS "owner",' +
           ' table_name AS "tableName", column_name AS "columnName",' +
           ' data_type AS "dataType",' +
-          ' character_maximum_length AS "length",' +
-          ' character_octet_length AS "dataLength",' +
+          ' character_maximum_length AS "dataLength",' +
           ' numeric_precision AS "dataPrecision",' +
           ' numeric_scale AS "dataScale",' +
           ' is_nullable = \'YES\' AS "nullable"' +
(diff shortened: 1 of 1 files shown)


slnode:

lib/discovery.js
```diff
@@ -144,6 +144,7 @@ function mixinDiscovery(MySQL) {
       sql = paginateSQL('SELECT table_schema AS "owner",' +
           ' table_name AS "tableName", column_name AS "columnName",' +
           ' data_type AS "dataType",' +
+          ' character_maximum_length AS "length",' +
           ' character_octet_length AS "dataLength",' +
```
raymondfeng: Can we just replace `dataLength` with `character_maximum_length`?

lib/discovery.js
```diff
@@ -156,6 +157,7 @@ function mixinDiscovery(MySQL) {
       sql = paginateSQL('SELECT table_schema AS "owner",' +
           ' table_name AS "tableName", column_name AS "columnName",' +
           ' data_type AS "dataType",' +
+          ' character_maximum_length AS "length",' +
```
raymondfeng: Same as above.

raymondfeng:
@partap Any chance to make the change as I suggested?

partap:
@raymondfeng 
Sorry, I left a note in the other issue...

I didn't want to replace `dataLength` because it could break software that may be relying on the current behavior of byte length.

partap:
@raymondfeng 
Hmm, I checked with a few different datatypes.  It seems that with blobs it does return the byte length value, so the only time it is different is when using a multibyte character set.  

So, it looks like the only backward compatibility issue _might_ be software that takes into account the extra space allocated for multibyte strings. Your call.

raymondfeng:
I think it's just a bug that we need to fix. We use the mysql.dataLength for the column length (in chars) during `automigrate`. The discovered result should match that.

partap:
@raymondfeng ok.

PR #81, opened 2015-03-08
Fix neq null

ulion:
Fix the "IS NOT NULL" query.


slnode:

PR #82, opened 2015-03-09
Do not change the source values for array in where parameter

ulion:
The connector should not change the input query parameters.

else if there is hardcoded query condition, when second time run it, it will not works because the array values are "escaped".

This PR fixed it.

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -310,10 +310,11 @@ MySQL.prototype.toDatabase = function (prop, val, forCreate) {
         this.toDatabase(prop, val[1]);
     } else if (operator === 'inq' || operator === 'nin') {
       if (Array.isArray(val)) { //if value is array
+        var vals = new Array(val.length);
         for (var i = 0; i < val.length; i++) {
-          val[i] = this.toDatabase(prop, val[i]);
+          vals[i] = this.toDatabase(prop, val[i]);
         }
-        return val.join(',');
+        return vals.join(',');
       } else {
         return this.toDatabase(prop, val);
       }


slnode:

lib/mysql.js
```diff
@@ -310,10 +310,10 @@ MySQL.prototype.toDatabase = function (prop, val, forCreate) {
         this.toDatabase(prop, val[1]);
     } else if (operator === 'inq' || operator === 'nin') {
       if (Array.isArray(val)) { //if value is array
-        for (var i = 0; i < val.length; i++) {
-          val[i] = this.toDatabase(prop, val[i]);
-        }
-        return val.join(',');
+        var self = this;
+        return val.map(function(v) {
```
raymondfeng: Please use a plain `for` loop to avoid performance penalty.

raymondfeng:
@ulion Please sign the CLA at https://cla.strongloop.com/agreements/strongloop/loopback-connector-mysql

ulion:
Code updated to not use map as you asked.

besides, the CLA ask to write to my github, I won't approve it because it looks like not necessary:
```
Review permissions

Repositories
Public only
This application will be able to read and write all public repo data. This includes the following:

Code
Issues
Pull requests
Wikis
Settings
Webhooks and services
Deploy keys
```

ulion:
Just updated to the fastest version, reference: http://jsperf.com/map-vs-native-for-loop/34

b-admike:
@ulion thank you for your contribution. The changes in this PR are also reflected in https://github.com/strongloop/loopback-connector-mysql/pull/94. I will close this one, so we can track it there.

slnode:

slnode:

slnode:

PR #87, opened 2015-03-15
Add test running instructions to readme

superkhau:
>+3. Set credentials somewhere?

@bajtos @raymondfeng @ritch Can someone clarify what the official instructions for this step is? (ie. set the username and password in ... file, hardcode key/value pair in `init.js` require('rc') empty object, etc)

Connected to strongloop/loopback#1167

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,3 +44,11 @@ To use it you need `loopback-datasource-juggler`.
     to `utf8_general_ci`. The `collation` value will also be used to derive the
     connection charset.
 
+## Running Tests
+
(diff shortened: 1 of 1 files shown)


bajtos:
@raymondfeng knows how the tests work. AFAIK, there is no init script to initialise your local database with data needed by tests, thus you have to put a .loopbackrc file to your project root to use our DB instance running in the cloud for the tests.

I really dislike this setup but don't have time to fix it.

superkhau:
@bajtos TY, I added steps to prevent future confusion. Merging without review as its a simple README update.

PR #90, opened 2015-03-24
Return count when updating or deleting models

superkhau:
- Add the `update/updateAll` function
- Add `deleteAll` alias to `destroyAll`

`dao.js` changes are required to make the API consistent across connectors (see strongloop/loopback-datasource-juggler#533)

Connect to strongloop/loopback#1167

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -641,6 +641,17 @@ MySQL.prototype.count = function count(model, callback, where) {
 
 };
 
+MySQL.prototype.update =
+MySQL.prototype.updateAll = function(model, where, data, cb) {
+  var query = 'UPDATE ' + this.tableEscaped(model) + ' SET ' +
+      this.toFields(model, data) + ' ' + this.buildWhere(model, where);
+  this.query(query, function(err, info) {
+    var affectedRows = info && typeof info.affectedRows === 'number' ?
+        info.affectedRows : undefined;
+    cb && cb(err, {count: affectedRows});
+  });
+};
+
 
 /**
  * Delete instances for the given model
@@ -657,8 +668,10 @@ MySQL.prototype.destroyAll = function destroyAll(model, where, callback) {
   }
   this.query('DELETE FROM '
       + this.tableEscaped(model) + ' ' + this.buildWhere(model, where || {}),
-    function (err, data) {
-      callback && callback(err, data);
+    function(err, info) {
+      var affectedRows = info && typeof info.affectedRows === 'number' ?
+          info.affectedRows : undefined;
+      callback && callback(err, {count: affectedRows});
     }.bind(this));
 };
 


superkhau:
@raymondfeng PTAL

lib/mysql.js
```diff
@@ -650,15 +669,23 @@ MySQL.prototype.count = function count(model, callback, where) {
...
       + this.tableEscaped(model) + ' ' + this.buildWhere(model, where || {}),
-    function (err, data) {
-      callback && callback(err, data);
+    function(err, info) {
+      // @superkhau The second arg is purposely set to an empty array for
+      // consistency across all connectors. The postgres connector currently
+      // returns `data.rows` (which is an array) in the second arg and cannot be
+      // changed for backwards compatibility reasons. So instead, we leave the
+      // second arg alone and pass metadata in the third arg instead.
+      callback && callback(err, [], {
```
bajtos: We need to preserve backwards compatibility of the mysql connector too! IMO you should keep passing `data` as the second callback argument.

I.e. any user code calling juggler and mysql-connector should keep receiving `data` in the callback.

lib/mysql.js
```diff
@@ -641,6 +642,24 @@ MySQL.prototype.count = function count(model, callback, where) {
...
+  var query = util.format('UPDATE %s SET %s %s', table, fields, whereClause);
+  this.query(query, function(err, info) {
+    // @superkhau The second arg is purposely set to an empty array for
+    // consistency across all connectors. The postgres connector currently
+    // returns `data.rows` (which is an array) in the second arg and cannot be
+    // changed for backwards compatibility reasons. So instead, we leave the
+    // second arg alone and pass metadata in the third arg instead.
+    cb && cb(err, [], {
+      count: info.affectedRows
+    });
```
bajtos: See my other comment, we need to preserve backwards compatibility of the mysql connector too. How was this implemented before? Was the function provided by juggler or was `updateAll` simply not working at all?

raymondfeng:
@superkhau Before your PRs, the `deleteAll` and `updateAll` only call back with `err`. There is no expectation to see the data. We should remove the `data` from the callback in your changes.

lib/mysql.js
superkhau: @bajtos `updateAll` was simply not implemented (it inherits from `loopback-connector`). I basically copy/pasted the implementation from `loopback-connector` so this should not break existing code as long as [put `info` back as the second arg](https://github.com/strongloop/loopback/issues/1167#issuecomment-85613965).

@raymondfeng I put the implementation to make things consistent (so we don't have a missing `updateAll` implementation in this connector only).

lib/mysql.js
```diff
@@ -650,15 +662,16 @@ MySQL.prototype.count = function count(model, callback, where) {
...
   if (!callback && 'function' === typeof where) {
     callback = where;
     where = undefined;
   }
   this.query('DELETE FROM '
       + this.tableEscaped(model) + ' ' + this.buildWhere(model, where || {}),
-    function (err, data) {
-      callback && callback(err, data);
+    function(err, info) {
+      callback && callback(err, {count: info.affectedRows});
```
superkhau: @raymondfeng Is this ok?

superkhau:
@raymondfeng Code review please.

bajtos:
>  Before your PRs, the deleteAll and updateAll only call back with err. There is no expectation to see the data. We should remove the data from the callback in your changes.

Just to clarify, there are two parts for this:

 - Juggler's updateAll is forwarding the second callback parameter passed by the connector implementation. We need to preserve this and pass the new "info" object as the third argument. See https://github.com/strongloop/loopback-datasource-juggler/blob/v2.21.0/lib/dao.js.

 - The `updateAll` implementation in the MySQL connector was forwarding the second query callback argument `results` (for whatever `result` being returned by `this.query`). We should preserve this too unless the `results` object was null/undefined.

raymondfeng:
@bajtos Before we try to enhance the updateAll/deleteAll with more feedback, even though the 2nd argument is passed into the callback function, it has always been `undefined`. No tests check or depend on the 2nd argument. That's why I feel that materializing the 2nd argument with count information is not breaking.

superkhau:
@raymondfeng PTAL again. I simplified the updates.

superkhau:
@slnode test please

lib/mysql.js
```diff
@@ -641,6 +641,15 @@ MySQL.prototype.count = function count(model, callback, where) {
 
 };
 
+MySQL.prototype.update =
+MySQL.prototype.updateAll = function(model, where, data, cb) {
+  var query = 'UPDATE ' + this.tableEscaped(model) + ' SET ' +
+      this.toFields(model, data) + ' ' + this.buildWhere(model, where);
+  this.query(query, function(err, info) {
+    cb && cb(err, {count: info.affectedRows});
```
raymondfeng: You need to check if info is set to avoid TypeError

lib/mysql.js
raymondfeng: Same as above

lib/mysql.js
```diff
@@ -641,6 +641,17 @@ MySQL.prototype.count = function count(model, callback, where) {
...
 };
 
+MySQL.prototype.update =
+MySQL.prototype.updateAll = function(model, where, data, cb) {
+  var query = 'UPDATE ' + this.tableEscaped(model) + ' SET ' +
+      this.toFields(model, data) + ' ' + this.buildWhere(model, where);
+  this.query(query, function(err, info) {
+    var affectedRows = info && typeof info.affectedRows === 'number' ?
+        info.affectedRows : undefined;
+    cb && cb(err, {count: affectedRows});
```
superkhau: @raymondfeng Fixed.

lib/mysql.js
```diff
@@ -657,8 +668,10 @@ MySQL.prototype.destroyAll = function destroyAll(model, where, callback) {
   }
   this.query('DELETE FROM '
       + this.tableEscaped(model) + ' ' + this.buildWhere(model, where || {}),
-    function (err, data) {
-      callback && callback(err, data);
+    function(err, info) {
+      var affectedRows = info && typeof info.affectedRows === 'number' ?
+          info.affectedRows : undefined;
+      callback && callback(err, {count: affectedRows});
```
superkhau: @raymondfeng Here too.

superkhau:
@raymondfeng PTAL again

raymondfeng:
LGTM

superkhau:
OK to merge from @raymondfeng on Slack.

PR #97, opened 2015-05-08
Refactor the code to use base SqlConnector

raymondfeng:
/to @ritch @bajtos 

This is a rewrite of the MySQL connector based on https://github.com/strongloop/loopback-connector/pull/16.

connect to strongloop/loopback-connector#14


raymondfeng:
connect to https://github.com/strongloop/loopback-connector/issues/14

lib/discovery.js
```diff
@@ -1,21 +1,46 @@
...
+ * @param {Object} mysql mysql driver
+ */
+function mixinDiscovery(MySQL, mysql) {
   var async = require('async');
 
   function paginateSQL(sql, orderBy, options) {
     options = options || {};
-    var limit = '';
+    var limitClause = '';
     if (options.offset || options.skip || options.limit) {
```
ritch: The `DAO` doesnt normalize this???

raymondfeng: Please note there is a story to refactor/clean up discovery/migration. This PR simply moves related code into their own files.

lib/mysql.js
```diff
@@ -127,21 +121,16 @@ require('util').inherits(MySQL, SqlConnector);
  * @param {String} sql The SQL statement
  * @param {Function} [callback] The callback after the SQL statement is executed
  */
-MySQL.prototype.query = function (sql, callback) {
+MySQL.prototype.executeSQL = function (sql, params, options, callback) {
```
ritch: This is confusing...? Can't we use `SQLConnector#execute()`. If this is mysql specific - maybe we should name the method so that is clear.

lib/mysql.js
```diff
@@ -374,929 +315,137 @@ MySQL.prototype._serializeObject = function(obj) {
...
-        case 'JSON':
-          break;
-        default:
-          if (!Array.isArray(props[p].type) && !props[p].type.modelName) {
-            // Do not convert array and model types
-            val = props[p].type(val);
-          }
-          break;
-      }
+MySQL.prototype.fromColumnValue = function(prop, val) {
```
ritch: I feel like the majority of this can be in the base connector. Thoughts?

lib/mysql.js
```diff
@@ -374,929 +315,137 @@ MySQL.prototype._serializeObject = function(obj) {
...
-  }
-  this.query('DELETE FROM '
-      + this.tableEscaped(model) + ' ' + this.buildWhere(model, where || {}),
-    function(err, info) {
-      var affectedRows = info && typeof info.affectedRows === 'number' ?
-          info.affectedRows : undefined;
-      callback && callback(err, {count: affectedRows});
-    }.bind(this));
+MySQL.prototype.getPlaceholderForIdentifier = function (key) {
+  return '??';
```
ritch: Would be nice if this was a constant, but might be a pre-opt.

lib/mysql.js
raymondfeng: We can push some them into the base SqlConnector as we already do for buildExpression.

lib/mysql.js
```diff
@@ -374,929 +315,137 @@ MySQL.prototype._serializeObject = function(obj) {
...
-      });
-    });
-  }, function(err) {
-    if (err) {
-      return err;
-    }
-    cb(null, !ok);
-  });
+MySQL.prototype.getCountForAffectedRows = function(model, info) {
+  var affectedRows = info && typeof info.affectedRows === 'number' ?
```
ritch: Is this really mysql specific? Maybe have this be the default impl in the baseconnector

lib/mysql.js
```diff
@@ -374,929 +315,137 @@ MySQL.prototype._serializeObject = function(obj) {
...
-  }
-  return dt;
-}
-
-function unsigned(p, dt) {
-  if (p.unsigned) {
-    dt += ' UNSIGNED';
-  }
-  return dt;
-}
```
ritch: +1 for all that code removed :) :+1:

test/migration.test.js
```diff
@@ -381,7 +381,7 @@ function setup(done) {
 }
 
 var query = function (sql, cb) {
-  db.adapter.query(sql, cb);
+  db.adapter.execute(sql, cb);
```
ritch: Why did this have to change? For consistency?

lib/discovery.js
ritch: ok

ritch:
LGTM :+1:

lib/mysql.js
raymondfeng: `executeSQL` is the extension point for the SqlConnector to implement the `execute` API, which does some normalization. See https://github.com/strongloop/loopback-connector/blob/feature/connector-refactor/lib/sql.js#L363-L389

lib/mysql.js
raymondfeng: Different DB drivers use different property names.

test/migration.test.js
raymondfeng: For consistency. It's in the direction to remove usage of aliases to favor the official method (`execute` over `query`).

lib/mysql.js
```diff
@@ -18,14 +18,48 @@ var debug = require('debug')('loopback:connector:mysql');
...
+    // loopback-datasource-juggler 2.x
+    modelBuilder.defineValueType.bind(modelBuilder) :
+    // loopback-datasource-juggler 1.x
+    modelBuilder.constructor.registerType.bind(modelBuilder.constructor);
+
+  // The Point type is inherited from jugglingdb mysql adapter.
+  // LoopBack uses GeoPoint instead.
+  // The Point type can be removed at some point in the future.
+  defineType(function Point() {
+  });
```
bajtos: An idea to consider: drop deprecated stuff like `Point` and support for loopback-datasource-juggler 1.x, and release the changes as a major version.

bajtos:
I did a quick superficial review, didn't find any obvious problems. FWIW, LGTM too.

PR #99, opened 2015-05-14
Add transaction support

raymondfeng:
/to @ritch @bajtos

This PR adds local transaction apis for relational DBs.

Connect to strongloop/loopback#109

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -26,7 +26,7 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
 
   dataSource.EnumFactory = EnumFactory; // factory for Enums. Note that currently Enums can not be registered.
 
-  process.nextTick(function () {
+  process.nextTick(function() {
     callback && callback();
   });
 };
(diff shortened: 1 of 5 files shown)


ritch:
Build is failing (probably due to the missing dep on transactions in juggler) but otherwise LGTM

heroca60:
Hi all.
Why loopback4 in the transactions retains and does not release the auto incremental id despite using await tx.rollback () when the transaction fails.
My repositories inherit from Default Transactional Repository as indicated in the official documentation.
for example..
...
Promise<Modulo> {
    //iniciando la transacción
    const tx: Transaction = await this.moduloRepository.beginTransaction({
      isolationLevel: IsolationLevel.SERIALIZABLE,
      timeout: 30000
    });
    let res: Modulo = new Modulo;
    try {
      res = await this.moduloRepository.create(modulo, {transaction: tx});
      await this.apartamentoRepository.updateById(modulo.idapartamento,
        {disponible: 1},
        {transaction: tx})
      await tx.commit();
    } catch (error) {
      await tx.rollback();
      throw new HttpErrors.Conflict('Ocurrió un problema. ' + error.message);
    }
    return res;
  }

...
The problem is that if the transaction fails, the self-generated id by mysql is not released, that is, for each failed transaction, an id is lost.

PR #113, opened 2015-07-21
Add support for RegExp operator

superkhau:
Connect to strongloop/loopback-datasource-juggler#552

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -459,6 +459,27 @@ MySQL.prototype.ping = function(cb) {
   this.execute('SELECT 1 AS result', cb);
 };
 
+MySQL.prototype.buildExpression = function(columnName, operator, operatorValue,
+    propertyDefinition) {
+  if (operator === 'regexp') {
+    if (operatorValue.ignoreCase)
+      console.warn('MySQL regex syntax does not respect the `i` flag');
+
+    if (operatorValue.global)
+      console.warn('MySQL regex syntax does not respect the `g` flag');
+
+    if (operatorValue.multiline)
+      console.warn('MySQL regex syntax does not respect the `m` flag');
+
+    return new ParameterizedSQL(columnName + ' REGEXP ?',
+        [operatorValue.source]);
+  }
+
+  // invoke the base implementation of `buildExpression`
+  return this.invokeSuper('buildExpression', columnName, operator,
+      operatorValue, propertyDefinition);
+};
+
 require('./migration')(MySQL, mysql);
 require('./discovery')(MySQL, mysql);
 require('./transaction')(MySQL, mysql);
--- a/package.json
+++ b/package.json
@@ -7,17 +7,18 @@
     "test": "mocha"
   },
   "dependencies": {
-    "loopback-connector": "^2.1.0",
-    "mysql": "^2.5.4",
     "async": "^0.9.0",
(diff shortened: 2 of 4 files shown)


lib/mysql.js
```diff
@@ -459,6 +459,22 @@ MySQL.prototype.ping = function(cb) {
   this.execute('SELECT 1 AS result', cb);
 };
 
+MySQL.prototype.buildExpression = function(columnName, operator, columnValue, propertyValue) {
+  if (operator === 'regexp' && typeof columnValue === 'string') {
```
raymondfeng: Assuming the normalization of the operand of `regexp` is done in juggler, we can clean up the code here.

superkhau:
@raymondfeng Review please.

lib/mysql.js
```diff
@@ -6,6 +6,7 @@ var mysql = require('mysql');
 var SqlConnector = require('loopback-connector').SqlConnector;
 var ParameterizedSQL = SqlConnector.ParameterizedSQL;
 var EnumFactory = require('./enumFactory').EnumFactory;
+var utils = require('loopback-datasource-juggler/lib/utils');
```
raymondfeng: This is invalid. `loopback-datasource-juggler` is a dev dep.

lib/mysql.js
```diff
@@ -459,6 +460,19 @@ MySQL.prototype.ping = function(cb) {
   this.execute('SELECT 1 AS result', cb);
 };
 
+MySQL.prototype.buildExpression = function(columnName, operator, columnValue,
+    propertyValue) {
+  if (operator === 'regexp') {
+    // strip regex flags because mysql doesn't allow them
+    var regex = utils.getRegExpExpression(columnValue);
```
raymondfeng: What does `utils.getRegExpExpression(columnValue);` do? `propertyValue` should already have the value of RegExp.

superkhau:
@raymondfeng Issues addressed.

lib/mysql.js
```diff
@@ -459,6 +459,16 @@ MySQL.prototype.ping = function(cb) {
   this.execute('SELECT 1 AS result', cb);
 };
 
+MySQL.prototype.buildExpression = function(columnName, operator, columnValue,
+    propertyValue) {
+  if (operator === 'regexp')
+    return new ParameterizedSQL(columnName + ' REGEXP ?', [columnValue.source]);
```
raymondfeng: Ignoring the `i` flag silently is not a good idea. We should either report the error or at least print out a warning.

superkhau:
@raymondfeng LGTY?

lib/mysql.js
```diff
@@ -459,6 +459,17 @@ MySQL.prototype.ping = function(cb) {
   this.execute('SELECT 1 AS result', cb);
 };
 
+MySQL.prototype.buildExpression = function(columnName, operator, operatorValue,
+    propertyDefinition) {
+  if (operator === 'regexp')
+    return new ParameterizedSQL(columnName + ' REGEXP ?',
```
raymondfeng: I believe we agreed to throw an error if the `i` flag is set.

superkhau:
@raymondfeng PTAL again, prints warnings this time. Also added tests to make sure warnings are being printed.

raymondfeng:
LGTM

PR #115, opened 2015-07-29
Fix RegExp unit test setup/teardown

superkhau:
Connect to strongloop/loopback-datasource-juggler#552


superkhau:
@raymondfeng Merging without review as its just minor setup/teardown refactoring.

PR #116, opened 2015-07-29
Add regexp operator tests

superkhau:
Add tests to ensure regexp operator works properly when flags are provided.


superkhau:
@raymondfeng Merging without review. Hardening with more tests again.

PR #117, opened 2015-07-29
Clean up regexop tests

superkhau:
Wrap long lines in regexp operator tests.


superkhau:
@raymondfeng Merging without review. Cleaning up long times in tests.

Issue #107, opened 2015-06-25
upsert() not working within transaction


nodejsbspa:
Hello!
I have found an issue when working with transaction and *upsert()* function. When executing the *upsert()* in a transaction, it blocks until the query timeouts.

It seems to me that *upsert()*, even if used within a transaction, doesn't use the same connection used by the other functions in the transaction but opens a new one. I have logged the connection id (*connection.threadId*) used by the functions and indeed the *upsert()* opens a new connection.

I checked the code and it looks to me that the function (line 210 of loopback-connector-mysql/lib/mysql.js)
```javascript
MySQL.prototype.updateOrCreate = MySQL.prototype.save =
  function(model, data, options, callback) {
....
}
```
take the *options* objects with transaction info in inputs but doesn't forward it to the execute method ( (line 237 of the same file):
```javascript
this.execute(sql.sql, sql.params, function(err, info){
   ....
``` 

I have changed line 237 of 'loopback-connector-mysql/lib/mysql.js' from 
```javascript
this.execute(sql.sql, sql.params, function(err, info) {
```
to
```javascript
this.execute(sql.sql, sql.params, options, function(err, info) { 
```
and this seems to resolve my issue: the connection id used by each function is the same.

Is it a real issue? Can it be fixed?
Thanks!

Amir-61:
It seems this issue is resolved by https://github.com/strongloop/loopback-connector-mysql/pull/128. Closing now... please feel free to re-open it if you see any issue.

Thanks!

PR #128, opened 2015-10-11
Fix database transactions not working with upsert()

devuo:
Implements the fix reported by @nodejsbspa in #107 in which the options passed to updateOrCreate were not passed to the execute method.


slnode:

PR #134, opened 2015-11-17
Add mysql CI host

cgole:


test/init.js
```diff
@@ -7,8 +7,8 @@ var config = require('rc')('loopback', {test: {mysql: {}}}).test.mysql;
 global.getConfig = function (options) {
 
   var dbConf = {
-    host: config.host || 'localhost',
-    port: config.port || 3306,
+    host: process.env.MYSQL_HOST || config.host || 'localhost',
+    port: process.env.MYSQL_PORT || config.port || 3306,
     database: 'myapp_test',
```
cgole: @rmg the database name is used by the tests so we'll just have to run those builds sequentially if they conflict. I'll add the username and password

rmg:
Would be nice if the tests were passing, but I'm guessing they were failing before this PR:
```
migrations
      ․ should run migration: 78ms
      ․ UserData should have correct columns: 5ms
      1) UserData should have correct indexes
      ․ StringData should have correct columns: 6ms
      ․ NumberData should have correct columns: 4ms
      ․ DateData should have correct columns: 3ms
      ․ should autoupdate: 117ms
      ․ should check actuality of dataSource: 24ms
      ․ should allow numbers with decimals: 23ms
      ․ should allow both kinds of date columns: 15ms
      ․ should report errors for automigrate: 0ms
      ․ should report errors for autoupdate: 0ms
      ․ should disconnect when done: 0ms
```

LGTM

test/init.js
```diff
@@ -3,15 +3,15 @@ module.exports = require('should');
...
-    username: config.username,
-    password: config.password,
+    username: process.env.MYSQL_USERNAME || config.username,
```
rmg: Should be `MYSQL_USER`, not `MYSQL_USERNAME`.

cgole: I can make that change, but may I ask why?

rmg: Because that is what is being set by CI and used by other repos that look for it.

cgole:
test please

test/init.js
```diff
@@ -3,15 +3,15 @@ module.exports = require('should');
 var DataSource = require('loopback-datasource-juggler').DataSource;
 
 var config = require('rc')('loopback', {test: {mysql: {}}}).test.mysql;
-
+console.log(config)
```
rmg: This should be removed before merging.

test/init.js
```diff
@@ -20,7 +20,7 @@ global.getConfig = function (options) {
       dbConf[el] = options[el];
     }
   }
-
+  console.log(dbConf)
```
rmg: This should be removed before merging.

rmg:
Other than the `console.log()` lines, LGTM.

cgole:
test please

cgole:
test please

rmg:
@cgole since the changes in this PR are only in the tests, you should be able to ignore any downstream/dependant build failures as they would be unrelated.

PR #139, opened 2015-12-08
Implementation of ReplaceOrCreate - mysql

Amir-61:
Hi @bajtos,

Please note this is just in progress implementation (Very initial implementation); just to make sure I'm on the right track.

##### Added functionality:
   +Implemented `replaceOrCreate` method for `loopback-connector-mysql`

##### Side note: 
* NPM test was failing out of the box.
* This patch requires changes from other modules including `loopback-datasource-juggler`, `loopback-connector` as well.

Connect to a spike task: https://github.com/strongloop-internal/scrum-loopback/issues/621

Thanks

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -200,17 +200,7 @@ MySQL.prototype.executeSQL = function(sql, params, options, callback) {
   }
 };
 
-/**
- * Update if the model instance exists with the same id or create a new instance
- *
- * @param {String} model The model name
- * @param {Object} data The model instance data
- * @param {Function} [callback] The callback function
- */
-MySQL.prototype.updateOrCreate = MySQL.prototype.save =
-  function(model, data, options, callback) {
-    var fields = this.buildFields(model, data);
-
+MySQL.prototype._modifyOrCreate = function(model, data, options, fields, cb) {
     var sql = new ParameterizedSQL('INSERT INTO ' + this.tableEscaped(model));
     var columnValues = fields.columnValues;
     var fieldNames = fields.names;
@@ -246,9 +236,36 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
         // 2 for each successful UPDATE.
         meta.isNewInstance = (info.affectedRows === 1);
       }
-      callback(err, data, meta);
+      cb(err, data, meta);
     });
   };
+  
+/**
+ * Replace if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [cb] The callback function
+ */
+MySQL.prototype.replaceOrCreate = function(model, data, options, cb) {
+  var fields = this.buildReplaceFields(model, data);
+  this._modifyOrCreate(model, data, options, fields, cb);
+};
+
+/**
+ * Update if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [cb] The callback function
+ */
+MySQL.prototype.save =
+MySQL.prototype.updateOrCreate = function(model, data, options, cb) {
+  var fields =  this.buildFields(model, data);  
+  this._modifyOrCreate(model, data, options, fields, cb);
+};
 
 function dateToMysql(val) {
   return val.getUTCFullYear() + '-' +
--- a/test/mysql.test.js
+++ b/test/mysql.test.js
@@ -156,6 +156,78 @@ describe('mysql', function () {
 
   });
 
+  context('replaceOrCreate', function() {
+    it('should replace the instance', function(done) {
+      Post.create({title: 'a', content: 'AAA'}, function(err, post) {
+        if (err) return done(err);
+        post = post.toObject();
+        delete post.content;
+        Post.replaceOrCreate(post, function(err, p) {
+          if (err) return done(err);
+          p.id.should.equal(post.id);
+          p.title.should.equal('a'); 
+          should.not.exist(p.content);
+          should.not.exist(p._id);
+          Post.findById(post.id, function(err, p) {
+            if (err) return done(err);
+            p.id.should.equal(post.id);
+            p.title.should.equal('a'); 
+            should.not.exist(post.content);
+            should.not.exist(p._id);
+            done();
+          });
+        });
+      });
+    });
+
+    it('should replace with new data', function(done) {
+      Post.create({title: 'a', content: 'AAA', comments: ['Comment1']},
+        function(err, post) {
+          if (err) return done(err);
+          post = post.toObject();
+          delete post.comments;
+          delete post.content;
+          post.title = 'b';
+          Post.replaceOrCreate(post, function(err, p) {
+            if (err) return done(err);
+            p.id.should.equal(post.id);
+            should.not.exist(p._id);
+            p.title.should.equal('b');        
+            should.not.exist(p.content);
+            should.not.exist(p.comments);
+            Post.findById(post.id, function(err, p) {
+              if (err) return done(err);
+              p.id.should.equal(post.id);
+              should.not.exist(p._id);
+              p.title.should.equal('b');
+              should.not.exist(p.content);
+              should.not.exist(p.comments);
+              done();
+            });
+          });
+        });
+    });
+
+    it('should create a new instance if it does not exist', function(done) {
+      var post = {id: 123, title: 'a', content: 'AAA'};
+      Post.replaceOrCreate(post, function(err, p) {
+        if (err) return done(err);
+        p.id.should.equal(post.id);
+        should.not.exist(p._id);
+        p.title.should.equal(post.title);
+        p.content.should.equal(post.content);
+        Post.findById(p.id, function(err, p) {
+          if (err) return done(err);
+          p.id.should.equal(post.id);
+          should.not.exist(p._id);
+          p.title.should.equal(post.title);
+          p.content.should.equal(post.content);
+          done();
+        });
+      });
+    });
+  });
+
   it('save should update the instance with the same id', function (done) {
     Post.create({title: 'a', content: 'AAA'}, function (err, post) {
       post.title = 'b';


lib/mysql.js
```diff
@@ -199,6 +199,55 @@ MySQL.prototype.executeSQL = function(sql, params, options, callback) {
...
 };
+ 
+/**
+ * Replace if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Function} [callback] The callback function
+ */
+
```
Amir-61: @bajtos: Maybe it makes more sense to have an argument in `MySQL.prototype.updateOrCreate` for `replaceOrCreate` and if value `true` is passed, we get the fields by calling `this.buildReplaceFields(model, data);` as opposed to `var fields = this.buildFields(model, data);so no need for duplicate code here, thoughts? However that requires a few changes where you were calling `updateOrCreate`,...

bajtos:
I don't see any obvious problems, besides the usual duplication of code.

@raymondfeng @superkhau PTAL too

bajtos:
@Amir-61 this comment of your got lost, I am posting it back from my email:

> Probably it makes more sense to have an argument in MySQL.prototype.updateOrCreate for replace and if value true is passed, we get the fields by calling this.buildReplaceFields(model, data); as opposed to var fields = this.buildFields(model, data); so no need for duplicate code here, thoughts?

We should definitely avoid code duplication. True/false arguments make the code difficult to read, can we perhaps pass the function to use (`this.buildFields` vs. `this.ReplaceFields` ) as the argument instead?

lib/mysql.js
```diff
@@ -249,6 +240,28 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+  
+/**
+ * Replace if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.replaceOrCreate = function (model, data, options, callback) {
+  this.modifyOrCreate(model, data, options, 'buildReplaceFields', callback);
```
Amir-61: @bajtos You proposed to use something like `this.buildReplaceFields` here; however the `buildReplaceFields` has `this` keyword in it; and it breaks `var props = this.getModelDefinition(model).properties;` inside `buildReplaceFields`; how about passing the function name like here and handeling it in `modifyOrCreate`?

bajtos: So there are two ways:

1) use the function name as you suggested, I think it's less clean because you cannot pass an anonymous function

2) pass a wrapper function that preserves `this`. You could use `this.buildReplaceFields.bind(this)`, except that `bind` has terrible performance. Therefore this would be a more performant solution:

```js
var self = this;
this.modifyOrCreate(model, data, options, function(model, data) { self.buildReplaceFields(model, data); }, callback);
```

Amir-61:
Hey Raymond,

Please also review this patch as well!

I really appreciate your time and feedback :-)

Amir-61:
@slnode test ok

Amir-61:
@slnode test please

lib/mysql.js
```diff
@@ -249,6 +240,30 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+/**
+ * Replace if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.replaceOrCreate = function (model, data, options, callback) {
+  var self = this;
+  this.modifyOrCreate(model, data, options, function(model, data) { return self.buildReplaceFields(model, data); }, callback);
```
Amir-61: Just clarifying: `buildReplaceFields` is a protected method implemented in `loopback-connector`; please see [this PR](https://github.com/strongloop/loopback-connector/pull/35)

Amir-61:
@superkhau/ @raymondfeng,

This is a small patch which is ready for review.

Thanks for your time!

Amir-61:
@superkhau / @raymondfeng,

This is a small patch which is ready for review.

I, also, just re-based and squashed the commits into one.

Thanks for your time!

lib/mysql.js
```diff
@@ -249,6 +240,30 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+/**
+ * Update if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.updateOrCreate = function (model, data, options, callback) {
+  var self = this;
+  this.modifyOrCreate(model, data, options, function(model, data) {return self.buildFields(model, data);}, callback);
```
superkhau: stick to 80 char limit, this one super long lin eis hard to read.

superkhau:
Can you add some tests to make sure we're not introducing regressions?

lib/mysql.js
```diff
@@ -249,6 +240,36 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.updateOrCreate = function (model, data, options, callback) {
+  var self = this;
+  this.modifyOrCreate(model, data, options, 
+    function(model, data) {return self.buildFields(model, data);}, callback);
```
superkhau: take out the inline function and assign it to a var in the line above

```
var someFunc = function(model, data) {
  ...
};
this.modifyOrCreate(model, data, options, someFunc, callback);
```

lib/mysql.js
```diff
@@ -249,6 +240,36 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.replaceOrCreate = function (model, data, options, callback) {
+  var self = this;
+  this.modifyOrCreate(model, data, options,
+    function(model, data) {
```
superkhau: move this up a line

lib/mysql.js
```diff
@@ -249,6 +240,36 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+      return self.buildReplaceFields(model, data);
```
superkhau: not sure if this return is necessary

lib/mysql.js
```diff
@@ -249,6 +240,36 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+
+/**
+ * Update if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.updateOrCreate = function (model, data, options, callback) {
```
superkhau: You need to add the save alias back.

```
MySQL.prototype.save =
MySQL.prototype.updateOrCreate = function(model, data, options, callback) {
...
```

test/mysql.test.js
```diff
@@ -155,7 +155,73 @@ describe('mysql', function () {
     });
 
   });
+
```
superkhau: context('replaceOrCreate', function() {
  it('should replace the instance', function(done) {
     ...
  });
  ...
});

lib/mysql.js
```diff
@@ -249,6 +240,39 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+
+/**
+ * Update if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.save = MySQL.prototype.updateOrCreate =
```
superkhau: you don't need to do this

just do what i posted in my last comment

```
MySQL.prototype.save =
MySQL.prototype.updateOrCreate = function ...
```

lib/mysql.js
```diff
@@ -249,6 +240,39 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+  function(model, data, options, callback) {
+    var self = this;
+    var fieldsGeneratorFunc = function(model, data) {
```
superkhau: can this function be moved into modifyOrCreate? does it need to be here?

Amir-61: No! `modifyOrCreate` is a shared function where you can pass function (which can be anonymous function) to build fields for either update or replace...

lib/mysql.js
```diff
@@ -200,17 +200,8 @@ MySQL.prototype.executeSQL = function(sql, params, options, callback) {
...
- *
- * @param {String} model The model name
- * @param {Object} data The model instance data
- * @param {Function} [callback] The callback function
- */
-MySQL.prototype.updateOrCreate = MySQL.prototype.save =
-  function(model, data, options, callback) {
-    var fields = this.buildFields(model, data);
-
+ MySQL.prototype.modifyOrCreate = function(model, data, options, buildFields ,callback) {
```
superkhau: should be `Fields, callback`

lib/mysql.js
```diff
@@ -200,17 +200,8 @@ MySQL.prototype.executeSQL = function(sql, params, options, callback) {
...
+    var fields = buildFields(model, data);
```
superkhau: Seem overly complicated passing in the functions. Can you just use options.buildType = 'replace/create' and then call `this.build[options.buildType]`. Then you remove both field generator functions.

superkhau: In fact, you can probably leave updateOrCreate alone. Then add a options.replace and check if (options.replace), then use this.buildReplaceFields instead of buildFields.

Amir-61: >Seem overly complicated passing in the functions.

Do think so? IMO it is not really complicated. I was following what you propose; please see [this comment](https://github.com/strongloop/loopback-connector-mysql/pull/139#discussion-diff-47161204)

superkhau: You probably understand the impl more than I do since you've had previous discussions with @bajtos. However, instead of passing in the function, can't you just pass in the fields?

MySQL.prototype.modifyOrCreate = function(model, data, options, fields, cb) {
  ...   
});
```

Then in updateOrCreate:

```
MySQL.prototype.updateOrCreate = function(model, data, options, callback) {
  var fields = this.buildReplaceFields(model, data);
  this.modifyOrCreate(model, data, options, fields, callback);
});
```

and in replace:

```
MySQL.prototype.replaceOrCreate = function (model, data, options, cb) {
  var fields = this.buildFields(model, data);
  this.modifyOrCreate(model, data, options, fields, callback);  
});
```

@raymondfeng Thoughts? ^

Comment (Amir-61) [inline lib/mysql.js]: :+1:

File: lib/mysql.js [hunk 14]
```diff
@@ -249,6 +239,35 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+ * Update if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.save =
+MySQL.prototype.updateOrCreate = function(model, data, options, callback) {
+  var self = this;
```
superkhau: don't need this line.

lib/mysql.js
```diff
@@ -249,6 +239,35 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+/**
+ * Replace if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.replaceOrCreate = function (model, data, options, callback) {
+  var self = this;
```
superkhau: don't need this line

lib/mysql.js
```diff
@@ -200,17 +200,7 @@ MySQL.prototype.executeSQL = function(sql, params, options, callback) {
...
- *
- * @param {String} model The model name
- * @param {Object} data The model instance data
- * @param {Function} [callback] The callback function
- */
-MySQL.prototype.updateOrCreate = MySQL.prototype.save =
-  function(model, data, options, callback) {
-    var fields = this.buildFields(model, data);
-
+ MySQL.prototype.modifyOrCreate = function(model, data, options, fields ,callback) {
```
superkhau: the comma is still in the wrong spot (there is a space in front of it, when there should only be one after)

lib/mysql.js
```diff
@@ -249,6 +239,35 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+  
+/**
+ * Replace if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [callback] The callback function
+ */
+MySQL.prototype.replaceOrCreate = function (model, data, options, callback) {
```
superkhau: remove space after function keyword

lib/mysql.js
```diff
@@ -249,6 +239,35 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
+  var self = this;
+  var fields = self.buildReplaceFields(model, data);
+  this.modifyOrCreate(model, data, options, fields, callback);
```
superkhau: can rename callback to cb

test/mysql.test.js
```diff
@@ -156,6 +156,75 @@ describe('mysql', function () {
 
   });
 
+  context('replaceOrCreate', function() {
+    it('should replace the instance', function (done) {
+      Post.create({title: 'a', content: 'AAA'}, function (err, post) {
```
superkhau: remove space in front of function

test/mysql.test.js
```diff
@@ -156,6 +156,75 @@ describe('mysql', function () {
...
+      var post = {id: 123, title: 'a', content: 'AAA'};
+      Post.replaceOrCreate(post, function (err, p) {
+        if (err) return done(err);
+        p.id.should.be.equal(post.id);
+        should.not.exist(p._id);      
+        p.title.should.be.equal(post.title);
+        p.content.should.be.equal(post.content);
+        Post.findById(p.id, function (err, p) {
+          if (err) return done(err);
+          p.id.should.be.equal(post.id);
```
superkhau: should.equal

lib/mysql.js
```diff
@@ -246,9 +236,36 @@ MySQL.prototype.updateOrCreate = MySQL.prototype.save =
...
     });
   };
+  
+/**
+ * Replace if the model instance exists with the same id or create a new instance
+ *
+ * @param {String} model The model name
+ * @param {Object} data The model instance data
+ * @param {Object} options The options
+ * @param {Function} [callback] The callback function
```
superkhau: cb

lib/mysql.js
```diff
@@ -200,17 +200,7 @@ MySQL.prototype.executeSQL = function(sql, params, options, callback) {
...
- * Update if the model instance exists with the same id or create a new instance
- *
- * @param {String} model The model name
- * @param {Object} data The model instance data
- * @param {Function} [callback] The callback function
- */
-MySQL.prototype.updateOrCreate = MySQL.prototype.save =
-  function(model, data, options, callback) {
-    var fields = this.buildFields(model, data);
-
```
superkhau: add some doc comments

test/mysql.test.js
```diff
@@ -156,6 +156,77 @@ describe('mysql', function () {
...
+            should.not.exist(post.content);
+            should.not.exist(p._id);
+            done();
+          });
+        });
+      });
+    });
+
+    it('should replace with new data', function(done) {
+      Post.create({title: 'a', content: 'AAA', comments: ['Comment1']}, function(err, post) {
```
superkhau: is this longer than 80 chars?

lib/mysql.js
Amir-61: But it is a private method; I just need to change it to `_modifyOrCreate` no?

superkhau: there you go ;)

superkhau:
LGTM

lib/mysql.js
```diff
@@ -200,17 +200,7 @@ MySQL.prototype.executeSQL = function(sql, params, options, callback) {
...
+ MySQL.prototype._modifyOrCreate = function(model, data, options, fields, cb) {
```
superkhau: there is an extra space in front of MySQL here

Amir-61:
@slnode test please

Amir-61:
@bajtos I prefer to wait for you to land this patch and this [PR](https://github.com/strongloop/loopback-connector/pull/35) since they are related and need to be landed at the same time.

PR #140, opened 2015-12-08
Fix ENUM test logic

superkhau:
This test should expect an error since it is adding the invalid ENUM
value "horse".

### DIFF
--- a/test/datatypes.test.js
+++ b/test/datatypes.test.js
@@ -8,9 +8,7 @@ describe('MySQL specific datatypes', function () {
   before(setup);
 
   it('should run migration', function (done) {
-    db.automigrate(function () {
-      done();
-    });
+    db.automigrate('EnumModel', done);
   });
 
   it('An enum should parse itself', function (done) {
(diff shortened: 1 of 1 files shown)


test/datatypes.test.js
```diff
@@ -39,13 +37,8 @@ describe('MySQL specific datatypes', function () {
...
   it('should fail spectacularly with invalid enum values', function (done) {
     var em = EnumModel.create({animal: 'horse', condition: 'sleepy', mood: 'happy'}, function (err, obj) {
-      assert.ok(!err);
-      EnumModel.findById(obj.id, function (err, found) {
-        assert.ok(!err);
-        assert.equal(found.animal, ''); // MySQL fun.
-        assert.equal(found.animal, 0);
-        done();
-      });
+      assert.ok(err);
```
bajtos: This will pass for any error (including connection failure). Could you please check that the test failed with an expected error? I believe MySQL provides error codes that can be used for that.

test/datatypes.test.js
```diff
@@ -37,15 +35,11 @@ describe('MySQL specific datatypes', function () {
...
-      EnumModel.findById(obj.id, function (err, found) {
-        assert.ok(!err);
-        assert.equal(found.animal, ''); // MySQL fun.
-        assert.equal(found.animal, 0);
-        done();
-      });
+  it('should fail with invalid enum values', function (done) {
+    EnumModel.create({animal: 'horse'}, function(err) {
+      assert.ok(err);
+      assert.equal(err.errno, 1265);
```
superkhau: @bajtos Fixed.

superkhau:
@bajtos PTAL again.

test/datatypes.test.js
superkhau: Actually this fails on ubuntu, but passes on mac. Have to dig deeper.

bajtos: Could you please add a comment explaining what does the magic number `1265` mean? Other than that, this LGTM, assuming it passes on all of our CI platforms.

bajtos: > Actually this fails on ubuntu, but passes on mac. Have to dig deeper.

My guess is that the error number may be MySQL-version specific?

superkhau:
Closing as this test is not relevant anymore with the juggler test updates (it already passes on master).

test/datatypes.test.js
superkhau: Yes this is MySQL-version specfiic, but I'm going to close this PR as juggler tests already pass on master.

PR #153, opened 2016-03-20
Allow migration to field types with no length

Kampfgnom:
Some MySQL field-types are being specified without a data length (e.g. BLOBs). Currently it is not possible to use such a custom column type in conjunction with the migrator since it will always try to create/modify a field like this:

```
    "icon": {
      "type": "string",
      "dataType": "LONGBLOB"
    }
```

with this MySQL statement, which is a syntax error:

```
ALTER TABLE `TableName` ADD COLUMN `icon` LONGBLOB(512) NULL
```

The proposed change will allow the author of a model configuration to add the field like this (note the dataLength: false):

```
    "icon": {
      "type": "string",
      "dataType": "LONGBLOB",
      "mysql": {
        "dataType": "LONGBLOB",
        "dataLength": false
      }
    }
```

This proposal is my first PR for a biggish OpenSource project and is in fact a request for help :) So if there is some better way to achieve what I am trying to do, you can happily reject his PR.

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -408,6 +408,9 @@ function mixinMigration(MySQL, mysql) {
     if (!prop) {
       return null;
     }
+    if (colType && columnMetadata && columnMetadata.dataLength === false) {
+      return colType;
+    }
     var colLength = columnMetadata && columnMetadata.dataLength ||
       prop.length || prop.limit;
     if (colType && colLength) {


slnode:

Amir-61:
@Kampfgnom Could you please write test to verify the changes?
Also please rebase your changes on top of master.

Thanks!

slnode:

slnode:

b-admike:
@Kampfgnom as @Amir-61 said, can you please rebase your branch against master and also write some unit tests?

slnode:

ssh24:
This issue should be fixed on the latest mysql release. It does not make sense to assign `dataLength` field a false value when it could be kept empty.

Please note: **BLOB types does not support data length, hence if you specify them in the `dataLength` it shall not work. Just keep the field empty.**

Closing this PR as a result.

PR #156, opened 2016-03-28
Use url to override other settings if url provided

jannyHou:
Connect to strongloop/loopback#2089

Use url to override other settings if url provided

```
if (s.url) {
   // Use url
} else {
  // Use other settings
}
```

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -72,31 +72,37 @@ function MySQL(settings) {
     s.connectionLimit = 10;
   }
 
-  var options = {
-    host: s.host || s.hostname || 'localhost',
-    port: s.port || 3306,
-    user: s.username || s.user,
-    password: s.password,
-    timezone: s.timezone,
-    socketPath: s.socketPath,
-    charset: s.collation.toUpperCase(), // Correct by docs despite seeming odd.
-    supportBigNumbers: s.supportBigNumbers,
-    connectionLimit: s.connectionLimit
-  };
+  var options;
+  if (s.url) {
+    // use url to override other settings if url provided
+    options = s.url;
+  } else {
+    options = {
+      host: s.host || s.hostname || 'localhost',
+      port: s.port || 3306,
+      user: s.username || s.user,
+      password: s.password,
+      timezone: s.timezone,
+      socketPath: s.socketPath,
+      charset: s.collation.toUpperCase(), // Correct by docs despite seeming odd.
+      supportBigNumbers: s.supportBigNumbers,
+      connectionLimit: s.connectionLimit
+    };
 
-  // Don't configure the DB if the pool can be used for multiple DBs
-  if (!s.createDatabase) {
-    options.database = s.database;
-  }
-
-  // Take other options for mysql driver
-  // See https://github.com/strongloop/loopback-connector-mysql/issues/46
-  for (var p in s) {
-    if (p === 'database' && s.createDatabase) {
-      continue;
+    // Don't configure the DB if the pool can be used for multiple DBs
+    if (!s.createDatabase) {
+      options.database = s.database;
     }
-    if (options[p] === undefined) {
-      options[p] = s[p];
+
+    // Take other options for mysql driver
+    // See https://github.com/strongloop/loopback-connector-mysql/issues/46
+    for (var p in s) {
+      if (p === 'database' && s.createDatabase) {
+        continue;
+      }
+      if (options[p] === undefined) {
+        options[p] = s[p];
+      }
     }
   }
 
--- a/test/connection.test.js
+++ b/test/connection.test.js
@@ -1,15 +1,40 @@
 require('./init.js');
 var assert = require('assert');
+var DataSource = require('loopback-datasource-juggler').DataSource;
+var mysqlConnector = require('../');
+var url = require('url');
 
-var db, DummyModel, odb;
+var db, DummyModel, odb, config;
 
 describe('connections', function () {
 
   before(function () {
     require('./init.js');
 
+    config = global.getConfig();
+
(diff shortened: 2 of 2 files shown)


jannyHou:
@bajtos please review it :) thank you

lib/mysql.js
```diff
@@ -72,31 +72,37 @@ function MySQL(settings) {
...
+      timezone: s.timezone,
+      socketPath: s.socketPath,
+      charset: s.collation.toUpperCase(), // Correct by docs despite seeming odd.
+      supportBigNumbers: s.supportBigNumbers,
+      connectionLimit: s.connectionLimit
+    };
+
+    // Don't configure the DB if the pool can be used for multiple DBs
+    if (!s.createDatabase) {
+      options.database = s.database;
```
bajtos: Are you sure this makes sense when `options` is a string as provided by `s.url`? The same question applies to L100-105.

bajtos:
@jannyHou please add a unit-test to verify the new feature.

bajtos:
> please add a unit-test to verify the new feature.

For example, create a datasource with both `url` and the old settings, where the old settings are invalid, and then make sure that the connector connects to the database specified by `url`.

lib/mysql.js
jannyHou: This snippet is inside `else` clause, not after, so it will not be executed if `options` is a string. 
I am aware of assigning var `options` a string value looks a little weird...I think change it to 
`mysql.createPool(url || options)`
would be more reasonable.

>>> For example, create a datasource with both `url` and the old settings, where the old settings are invalid, and then make sure that the connector connects to the database specified by `url`

As I explained in https://github.com/strongloop/loopback-connector-db2/pull/23#issuecomment-203496208
I verified this on my local for each connectors lol, then commit those PRs. I will think of a way to create the unit test for connection testing.

jannyHou:
@bajtos 
hmm...let me rephrase my concern about the unit test and keep our discussion in this repo, because only mysql contains a connection test script even though not exactly the test we expected lol:

In each connector repo's test folder, there is a file called [`init.js`](https://github.com/strongloop/loopback-connector-mysql/blob/master/test/init.js).
It reads the [config information from server env](https://github.com/strongloop/loopback-connector-mysql/blob/master/test/init.js#L9-L16)
Then declare global function [`getDataSource`](https://github.com/strongloop/loopback-connector-mysql/blob/master/test/init.js#L26-L28), so other test cases could call it to build the connection.

I have two directions about how to write the connection test case:
- [1]Setup url of every connector on every server

Honestly I am not quite familiar with setup env on jenkins hence it's hard to estimate the complexity and efforts we need to take.
Do you think this is a reasonable way or do we need to get more suggestions from ryan?
- [2] We have test config for [mongodb/mssql/mysql/postgresql/oracle](https://github.com/strongloop/strongloop-config/blob/master/.loopbackrc#L55) and [a temporary db2 config](https://w3-connections.ibm.com/wikis/home?lang=en-gb#!/wiki/Wc226060fbeac_49d3_a3ee_a5756a5ef00a/page/VM%20Server%20Info)

My concern is are they stable enough(especially the db2) to use from the perspective of long-term maintain?


And from those connectors supporting url, no tests for connection seem exist, so I think no matter which method we use, we'd better create a new file called 'connection.test.js'(like in mysql) 
Any thoughts?

jannyHou:
@slnode test please

bajtos:
@jannyHou AFAICT, the `test/init.js` file exports also global `getConfig` method that returns the config object used by `getDataSource`, at least in this repository.

So in order to test this functionality, we can build the URL from the values provided by `getConfig`. I am thinking about something like this:

```js
var validConfig = getConfig();
var url = // build from validConfig
var dsConfig = {
  url: url,
  host: 'invalid-hostname',
  port: 80, // this is HTTP, no MySQL server should listen there
  database: 'invalid-database',
  username: 'invalid-username',
  password: 'invalid-password'
};

var ds = // create datasource using dsConfig
ds.ping(done); // done is the callback of the test method
```

Thoughts?

jannyHou:
Thank you @bajtos smart solution! And it's a reasonable solution for all connectors.

jannyHou:
@bajtos I added unit test for connection, I think they pass the CI test 
>  should pass with valid settings: 369ms
      ․ shoud pass with valid URL override invalid settings: 267ms

The test fails due to hook testing(not related to this pr I believe), I am looking into those failures now, will fix them.
Please review it. Thanks :)

test/connection.test.js
```diff
@@ -1,17 +1,42 @@
 require('./init.js');
 var assert = require('assert');
+var DataSource = require('loopback-datasource-juggler').DataSource;
+var mysqlConnector = require('../');
+var URL = require('url');
```
bajtos: Nitpick: we usually use the same property name as is the module name, i.e. lowercase `url`.

test/connection.test.js
```diff
@@ -1,17 +1,42 @@
...
     odb = getDataSource({collation: 'utf8_general_ci', createDatabase: true});
     db = odb;
   });
 
+  it('should pass with valid settings', function (done) {
+    var db = new DataSource(mysqlConnector, config);
+    db.ping(done);
+  });
+
+  it('shoud pass with valid URL override invalid settings', function (done) {
```
bajtos: I think this may be better phrasing: `(it) should pass when valid URL overrides invalid settings`.

On the second thought, the test name should be ideally focused on high-level user perspective, in which case this may be a better name: `(it) ignores all other settings when url is present`.

bajtos:
Two minor details to improve or ignore, as you wish. No further review is needed, please squash the commits before landing.

PR #160, opened 2016-04-09
Allow multipleStatements connection setting

cvrabie:
Allows you to set the `multipleStatements` mysql connection setting. 

Based on old forum [posts](https://groups.google.com/forum/#!topic/loopbackjs/3dZBUa9ZXWo) this seems to have worked at some point. However it was disabled in the mysql [library](https://github.com/felixge/node-mysql/blob/master/Readme.md#multiple-statement-queries) to help with prevention of sql injection.

This small patch is coded in such a way that it preserves the default, whichever that may be, in your version of the mysql library. It does however allow you to override it if you know what you're doing.

This is particularly important if you're using stored procedures as the most common way to get data out of a call is something like:

``` sql
CALL stored_proc(in_1, in_2, @out_1, @out_2); SELECT @out_1, @out_2;
```

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -104,6 +104,10 @@ function MySQL(settings) {
         options[p] = s[p];
       }
     }
+
+    if(typeof s.multipleStatements === 'boolean'){
+      options.multipleStatements = s.multipleStatements;
+    }
   }
 
   this.client = mysql.createPool(options);


slnode:

lib/mysql.js
```diff
@@ -104,6 +104,10 @@ function MySQL(settings) {
         options[p] = s[p];
       }
     }
+
+    if(typeof s.multipleStatements === 'boolean'){
+      options.multipleStatements = s.multipleStatements;
+    }
```
raymondfeng: Isn't it already covered by line 104?

ssh24:
@cvrabie Could you please rebase your branch? Also please add some unit tests to go with this patch.

slnode:

slnode:

slnode:

jannyHou:
@cvrabie agree with Raymond's comment https://github.com/strongloop/loopback-connector-mysql/pull/160#r74452276 it's already set in LINE104, PTAL if you are still working on it. Thanks.

cvrabie:
Sorry guys, I haven't worked with this in a long time so I don't remember anything about it other than it was not working at the time. However, looking at it now, it seems Raymond is right and it should be covered by the above line. This pull request seems unnecessary. I will close it.

PR #166, opened 2016-05-09
Add function connect

jannyHou:
Connect to strongloop/loopback#2242

Create function `connect` in mysql connector.
When booting app from swagger generator, datasources are setup with ds.lazyConnect = true to avoid connection.

Still work in progress

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -31,9 +31,15 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
 
   dataSource.EnumFactory = EnumFactory; // factory for Enums. Note that currently Enums can not be registered.
 
-  process.nextTick(function() {
-    callback && callback();
-  });
+  if(callback) {
+    if(dataSource.settings.lazyConnect) {
+      process.nextTick(function() {
+        callback();
+      });
+    } else{
+      dataSource.connector.connect(callback);
+    }
+  }
 };
 
 exports.MySQL = MySQL;
@@ -60,7 +66,41 @@ function defineMySQLTypes(dataSource) {
  */
 function MySQL(settings) {
   SqlConnector.call(this, 'mysql', settings);
+}
 
+require('util').inherits(MySQL, SqlConnector);
+
+MySQL.prototype.connect = function(callback) {
+  var self = this;
+  var options = generateOptions(this.settings);
+  var s = self.settings || {};
+
+  if (this.client) {
+    if (callback) {
+      process.nextTick(function() {
+        callback(null, self.client);
+      });
+    }
+  } else {
+    this.client = mysql.createPool(options);
+    this.client.getConnection(function(err, connection) {
+      var conn = connection;
+      if (!err) {
+        if (self.debug) {
+          debug('MySQL connection is established: ' + self.settings || {});
+        }          
+        connection.release();
+      } else {
+        if (self.debug || !callback) {
+          console.error('MySQL connection is failed: ' + self.settings || {}, err);
+        }
+      }
+      callback && callback(err, conn);
+    });
+  }
+}
+
+function generateOptions (settings) {
   var s = settings || {};
   if (s.collation) {
     // Charset should be first 'chunk' of collation.
@@ -110,22 +150,8 @@ function MySQL(settings) {
       }
     }
   }
-
-  this.client = mysql.createPool(options);
-
-  this.client.on('error', function(err) {
-    dataSource.emit('error', err);
-    dataSource.connected = false;
-    dataSource.connecting = false;
-  });
-
-  if (debug.enabled) {
-    debug('Settings: %j', s);
-  }
+  return options;
 }
-
-require('util').inherits(MySQL, SqlConnector);
-
 /**
  * Execute the sql statement
  *
@@ -250,7 +276,7 @@ MySQL.prototype._modifyOrCreate = function(model, data, options, fields, cb) {
       cb(err, data, meta);
     });
   };
-  
+
 /**
  * Replace if the model instance exists with the same id or create a new instance
  *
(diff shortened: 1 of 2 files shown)


bajtos:
IIRC, loopback-connector-mysql uses connection pooling, that's why there is no `connect` called. Perhaps we can defer creation of the connection pool when `lazyConnect:true`? Having said that, I have no idea whether such change will actually change any observable behaviour.

jannyHou:
@raymondfeng I am aware of that mysql connector doesn't have `connect` phase in it, and the invalid config would not throw or emit error when `initialize`. From my observation, run an app (`node .`) or run `apic loopback:refresh` with an invalid mysql config would not exit with error.

Do you think I should create a `connect` phase in mysql? Any thoughts?
Thanks!

raymondfeng:
MySQL driver uses `createPool` to create a connection pool (https://github.com/strongloop/loopback-connector-mysql/blob/feature/lazy-connect/lib/mysql.js#L120). We should wrap it into `connect` instead.

jannyHou:
@raymondfeng I separate the `connect` phase from MySQL constructor:
https://github.com/strongloop/loopback-connector-mysql/blob/master/lib/mysql.js#L114-L124

And call `mysql.client.getConnection(settings)` to test the connection. 

Please review. Thanks!

lib/mysql.js
```diff
@@ -60,7 +66,39 @@ function defineMySQLTypes(dataSource) {
...
+
+  if (this.client) {
+    if (callback) {
+      process.nextTick(function() {
+        callback(null, self.client);
+      });
+    }
+  } else {
+    this.client = mysql.createPool(options);
+    this.client.getConnection(function(err, connection) {
```
raymondfeng: Why do you try to get a connection? To test out the pool?

lib/mysql.js
```diff
@@ -60,7 +66,39 @@ function defineMySQLTypes(dataSource) {
...
+      if (!err) {
+        if (self.debug) {
+          debug('MySQL connection is established: ' + self.settings || {});
+        }
+      } else {
+        if (self.debug || !callback) {
+          console.error('MySQL connection is failed: ' + self.settings || {}, err);
+        }
+      }
+      callback && callback(err, connection);
```
raymondfeng: The connection should be released if it's for testing purpose.

lib/mysql.js
jannyHou: @raymondfeng Yeah, the thing is if I use the original code, no `error` event would be triggered:
```js
this.client = mysql.createPool(options);

  this.client.on('error', function(err) {
    dataSource.emit('error', err);
    dataSource.connected = false;
    dataSource.connecting = false;
  });
```

So I have to call `getConnection` to get the connection error message.

lib/mysql.js
jannyHou: Fixed.
I release the connection after it's established.

lib/mysql.js
raymondfeng: OK. You can call `ping()` too.

raymondfeng:
LGTM

andrs1294:
Hi guys @jannyHou @raymondfeng , 
I think that there is a problem with this. 
Before, can you read [this](https://github.com/strongloop/loopback/issues/2242#issuecomment-331651873) please ??

So, When `apic loopback:refresh` try to get the swagger file, The connection is skipped, So, I that example, The boot script try to automigrate the data, This will crash because it can't drop the table in the database.

The error happens in loopback-connector-mysql/lib/mysql.js - Line  455
`MySQL.prototype.escapeName = function(name) {
  return this.client.escapeId(name);
};`

**The this.client is undefined.** So the app crashed when I try to get the swagger file with the example.

This is the stack of the error.
> TypeError: Cannot read property 'escapeId' of undefined
    at MySQL.escapeName ({{PROYECT}}node_modules/loopback-connector-mysql/lib/mysql.js:456:21)
    at MySQL.SQLConnector.tableEscaped ({{PROYECT}}node_modules/loopback-connector-mysql/node_modules/loopback-connector/lib/sql.js:232:15)
    at MySQL.SQLConnector.dropTable ({{PROYECT}}node_modules/loopback-connector-mysql/node_modules/loopback-connector/lib/sql.js:1294:47)
    at {{PROYECT}}node_modules/loopback-connector-mysql/node_modules/loopback-connector/lib/sql.js:1341:10
    at {{PROYECT}}node_modules/loopback-connector-mysql/node_modules/loopback-connector/node_modules/async/lib/async.js:181:20
    at Object.async.forEachOf.async.eachOf ({{PROYECT}}node_modules/loopback-connector-mysql/node_modules/loopback-connector/node_modules/async/lib/async.js:233:13)
    at Object.async.forEach.async.each ({{PROYECT}}node_modules/loopback-connector-mysql/node_modules/loopback-connector/node_modules/async/lib/async.js:209:22)
    at MySQL.SQLConnector.automigrate ({{PROYECT}}node_modules/loopback-connector-mysql/node_modules/loopback-connector/lib/sql.js:1340:9)
    at DataSource.automigrate ({{PROYECT}}node_modules/loopback-datasource-juggler/lib/datasource.js:964:18)
    at createCoffeeShops ({{PROYECT}}server/boot/create-sample-models.js:35:13)
    at {{PROYECT}}node_modules/async/dist/async.js:1268:19
    at {{PROYECT}}node_modules/async/dist/async.js:3853:24
    at replenish ({{PROYECT}}node_modules/async/dist/async.js:946:17)
    at {{PROYECT}}node_modules/async/dist/async.js:950:9
    at eachOfLimit ({{PROYECT}}node_modules/async/dist/async.js:975:24)
    at {{PROYECT}}node_modules/async/dist/async.js:980:16

Thanks, have a nice day. Excuse for my english =)

PR #169, opened 2016-06-02
fix a bug for explicit column dataType definition in automigration/au…

pluswave:
the model definition :

"modelName": {
  "type" : "someType",
  "mysql" : {
   "dataType": "someOtherType"
  }
}
when doing auto migration / auto update, it should use someOtherTye instead of
the default map of someType.
however, there was a bug that missed the 'mysql' level for intepreting this.
this commit fixed this bug.

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -729,8 +729,8 @@ function mixinMigration(MySQL, mysql) {
 
   function columnType(p, defaultType) {
     var dt = defaultType;
-    if (p.dataType) {
-      dt = String(p.dataType);
+    if (p.mysql && p.mysql.dataType) {
+      dt = String(p.mysql.dataType);
     }
     return dt;
   }


slnode:

lib/migration.js
```diff
@@ -461,8 +461,8 @@ function mixinMigration(MySQL, mysql) {
 
   function columnType(p, defaultType) {
     var dt = defaultType;
-    if (p.dataType) {
-      dt = String(p.dataType);
```
rlloyd2001: I'm worried about backwards compatibility here... Maybe do an "else if" keeping this code.

ssh24: Agree with @rlloyd2001. Please put this on a if statement and unify the logic.

ssh24:
@slnode ok to test

slnode:

slnode:

slnode:

ssh24:
@pluswave Please rebase your PR and fix the commit linter error
```
**
**  Linting commit logs
**
**  3 problems found:
**    3abdcf7 - fix a bug for explicit column dataType definition i: First line should be 50 characters or less (saw 78)
**    3abdcf7 - fix a bug for explicit column dataType definition i: line 11 longer than 72 characters.
**    3abdcf7 - fix a bug for explicit column dataType definition i: line 13 longer than 72 characters.
**
```

lib/migration.js
pluswave: but 'else if' is not compatible with doc.

pluswave:
@ssh24  I think the lint error is not introduced by this PR.

ssh24:
@pluswave It is on this PR. Your commit messages are too long. The first line of your commit should be less than 50 characters.

ssh24:
@pluswave Could you please rebase your branch or give me the access to your fork repo so that I can rebase it for you? https://github.com/blog/2247-improving-collaboration-with-forks

pluswave:
@ssh24 , I have rebased my branch, but seems there are some other issues. How can I invite you to edit my branch ? seems the link is about new PR but not old one.

ssh24:
@pluswave You should have a checkbox  on the side saying `Allow edits from maintainers`. Please check that to give me the access.

pluswave:
@ssh24  ok I just saw it and checked.

lib/migration.js
ssh24: @pluswave This affects the backwards compatibility here. Breaking a bunch of tests related to it. I agree with your changes but it should go with the purpose of the fix. 

Let's say we have the model definition:
```
"modelName": {
"type" : "someType",
"mysql" : {
"dataType": "someOtherType"
}
}
```
**When doing auto migration / auto update, it should use someOtherType instead of
the default map of someType.**

With your patch, it will try to always look for `mysql.OtherType` and if it does not exists it would just `return dt;`, not acceptable because the type could have been also set by `"type": "someType"`.

Hence, I would add some tweaks to accept both ways, something like this:
```
    if (p.mysql && p.mysql.dataType) {
      dt = String(p.mysql.dataType);
    }
    if (p.dataType) {
      dt = String(p.dataType);
    }
```

pluswave: for backwards compatible,  I pick a case here for discussion:

```
"modelName": {
  "type" : "someType",
  "dataType": "someDataType",
  "mysql" : {
   "dataType": "someOtherType"
  }
}
```
what column name should be used in the database ? IMHO, it should be "someOtherType", because it is defined in mysql and it is expected to be.  We should introduce new test(s) or change existing test(s) to make it clear.

ssh24: I completely agree with you. Please add/change new/existing test cases in this PR as part of this patch.

ssh24:
@pluswave Any updates on the comments I left and the fixes required?

ssh24:
Closing this PR since there has been no response on it. This is definitely a bug so it is taken note of in our backlog. #281

PR #170, opened 2016-06-03
fix and refactoring timestamp and datetime data types handling

darknos:
Fix issue with values like '0000-00-00 00:00:00' and prevent twice converted timezones.

datasource.options.timezone and datasource.options.dateStrings now work correct.

PR #179, opened 2016-07-26
Add special handling of zero date/time entries

azatoth:
Per MySQL docs
(http://dev.mysql.com/doc/refman/5.7/en/date-and-time-types.html):

  "MySQL permits you to store a “zero” value of '0000-00-00' as a
  “dummy date.” This is in some cases more convenient than using NULL
  values, and uses less data and index space. To disallow '0000-00-00',
  enable the NO_ZERO_DATE mode.

  “Zero” date or time values used through Connector/ODBC are converted
  automatically to NULL because ODBC cannot handle such values."

As we are not using Connector/ODBC we need to handle this ourself.

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -405,7 +405,15 @@ MySQL.prototype.fromColumnValue = function(prop, val) {
         val = String(val);
         break;
       case 'Date':
-        val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
+
+        // MySQL allows, unless NO_ZERO_DATE is set, dummy date/time entries
+        // new Date() will return Invalid Date for those, so we need to handle
+        // those separate.
+        if (val == '0000-00-00 00:00:00') {
+          val = null;
+        } else {
+          val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
+        }
         break;
       case 'Boolean':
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

azatoth:
While PR #170 contains a fix for this issue, I made this PR to only contain the fix and no other refactorings in hope it has a better chance of getting accepted. I mean no disrespect to @darknos's contribution, and are fully OK his PR gets accepted instead of this one.

raymondfeng:
@azatoth Thanks for the patch. Can you add a test case to cover the special case?

azatoth:
@raymondfeng I would love to but I'm uncertain how to setup the database for executing tests; README.md says I should ask a core developer for instructions.
Also I assume proper location for test is in migration.test.js?

superkhau:
@azatoth Just take a look at https://github.com/strongloop/loopback-connector-mysql/blob/master/test/init.js#L10 and change the configs for your local MYSQL instance. We definitely need to work on making this info more public and easily digestable for the community.

azatoth:
@raymondfeng I've added a test for this issue now; I hope it's sufficient

rmg:
@slnode test please

PR #184, opened 2016-08-05
Update tests to allow concurrency and newer MySQL

rmg:
The PR finishes up the work started on in #134 and expands it to support all DB connection details. Additionally, it replaces the use of the globally shared mysql test DB from strongloop-config with a pre-test action that populates the local database accordingly. With this change, multiple instances of the test suite can be run concurrently as long as they are running on different hosts or have environment variables configured to point at different database servers.

During development, I noticed that several of the tests themselves were relying on behaviour which has changed between MySQL 5.5 and MySQL 5.7. Some of these differences have been accounted for by checking the version and altering the expectation and others have been modified to skip certain tests.


rmg:
@slnode test please

rmg:
* downstream test for `multitenancy@master` is failing due to strongloop/loopback-multitenant-poc#4
* downstream test for `loopback-connector-mysql-jdbc@mater` is failing because cis-jenkins only has a JRE installed

rmg:
@slnode test please

rmg:
@strongloop/loopback-maintainers please take a look. This PR should allow the tests for this module to be run in parallel, which should reduce test time for many upstream modules.

@raymondfeng would you mind taking a look at the discovery test I marked for skipping? I believe this might actually be a long-standing bug in this module, since [lib/discovery.js:77](https://github.com/strongloop/loopback-connector-mysql/blob/0fcaa35f52ca1963655ae2a4caaa49e5d4355fe9/lib/discovery.js#L77) makes the assumption that database/schema maps to user name.

test/mysql.discover.test.js
```diff
@@ -74,7 +75,8 @@ describe('discoverModels', function () {
   });
 
   describe('Discover models excluding views', function () {
-    it('should return an array of only tables', function (done) {
+    // TODO: this test assumes the current user owns the tables
+    xit('should return an array of only tables', function (done) {
```
0candy: typo extra `x`

superkhau:
LGTM

test/mysql.discover.test.js
rmg: `xit` is an alias for `it.skip`.. I guess it's only obvious if you're familiar with the convention. I'll change it to be more obvious.

0candy: ok LGTM

loay:
LGTM

deepakrkris:
@rmg verified and tested locally. Please go ahead.

deepakrkris:
@rmg you are skipping the “discover models excluding views” but keeping the “discover models including views” , why is that ?

rmg:
@deepakrkris that is why I asked @raymondfeng to take a look at:
> would you mind taking a look at the discovery test I marked for skipping? I believe this might actually be a long-standing bug in this module, since lib/discovery.js:77 makes the assumption that database/schema maps to user name.

It isn't about finding or not finding views. The problem is triggering the default code path for discovery which looks for all tables/views owned by a user, but they are actually owned by a parent database/schema. I think the only reason it was passing before was because the db username happened to be the same as the database name.

deepakrkris:
@rmg if schema ownership is the issue, it would apply for both the test cases, discover including views and excluding views

pretest.js
```diff
@@ -0,0 +1,28 @@
...
+}
+if (process.env.TEST_MYSQL_PASSWORD) {
+  args.push('--password=' + process.env.TEST_MYSQL_PASSWORD);
+}
+
+console.log('seeding DB with example db...');
+var mysql = cp.spawn('mysql', args, {stdio: stdio});
+sql.pipe(mysql.stdin);
+mysql.on('exit', function() {
+  console.log('done seeding DB');
```
bajtos: @rmg How are you handling errors? I see that mysql's stderr is forwarded to stderr, so at least we will see any error messages in console logs.

Should you check for the exit code of the `mysql` process and print a different error message & return non-zero exit code if the `mysql` process failed?

bajtos: @rmg Feel free to address this in a follow-up pull request, so that you can land this big improvement sooner.

PR #186, opened 2016-08-09
test: use original test db as seed schema

rmg:
Followup to #184 to debug environment on ci.strongloop.com.

It turns out that a bunch of the weird case inconsistencies were actually just that the example DB looked _a lot_ like the test db at first glance. This PR takes a proper dump of the `STRONGLOOP` database from the shared test server and uses _it_ as the seed data, reverting some of the case changes from #184.

After *this* PR is merged we can try again to make the builds for this repo run in parallel, which should give ~10x faster builds (5 node versions * 2 linux flavours).

### DIFF
--- a/pretest.js
+++ b/pretest.js
@@ -13,7 +13,7 @@ if (!process.env.TEST_MYSQL_USER) {
 var fs = require('fs');
 var cp = require('child_process');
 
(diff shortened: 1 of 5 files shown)


superkhau:
LGTM

Issue #135, opened 2015-11-25
Add auto creation of foreign keys to autoupdate and automigrate


kiraLinden:
See https://github.com/strongloop/loopback-connector-mysql/issues/16 automigate and autoupdate currently don't create foreign keys, and it would be nice to have for referential integrity.

XinV:
Up vote for this issue, it seems against the claim of serving as an enterprise framework if not having such referential integrity in place @raymondfeng @bajtos

bajtos:
+1

nautskye:
Yes please!

remkohdev:
Is this still the case or am I misreading this? that when I have a 'built-in User' model with a relation 'hasMany' to my custom 'myObject', which works like a charm in memory db, that this relation is not 'automigrated' to a relational DB like PostGres, but that I have to code the extra foreign key field and maintain the relationship by explicitly saving it in my create/update and add it possibly to my delete and find filters? 
It would be super nice if it would automigrate it ofcourse, but the least that could be done, is document this properly perhaps?

kiraLinden:
There's a way to manually do it https://apidocs.strongloop.com/loopback-datasource-juggler/#datasource-prototype-defineforeignkey but having it done without having to do that extra step would be nice.

antonsamper:
@kiraLinden I have tried using `dataSource.defineForeignKey(className, key, foreignClassName, pkName)` but this doesn't explicitly set a field as a foreign key in the database.

kiraLinden:
Oh sorry. This should probably be looked at by the loopback team...

pluswave:
+1

ferryhinardi:
+1

antonsamper:
@bajtos do you know if there has been any progress with this issue?

bajtos:
@antonsamper no progress so far. Would you like to contribute this feature yourself? We can help you along the way.

madianas:
+1

matthewdickinson:
@bajtos I would be willing to take a stab at implementing this. Is there an existing connector that implements this that I should review as an example, or would this be the first connector to do this?

bajtos:
> > See #16 automigate and autoupdate currently don't create foreign keys, and it would be nice to have for referential integrity.

> I would be willing to take a stab at implementing this. Is there an existing connector that implements this that I should review as an example, or would this be the first connector to do this?

[...]

PR #193, opened 2016-09-19
Foreign key migration

matthewdickinson:
@superkhau Here is a beginning to supporting foreign keys in autoupdate/automigrate (per https://github.com/strongloop/loopback-connector-mysql/issues/135).

This does build and drop the database side foreign keys when foreign keys are defined on the model. (I have a [fork of the datasource juggler](https://github.com/matthewdickinson/loopback-datasource-juggler/commits/foreign_key_generation) that exposes the foreign keys via the `settings` for the model, but I wasn't sure what the best way to connect that is.)

These are the items that I know could happen/be issues
- Trying to build a foreign key across datasources (I haven't yet added anything to bar a user from creating a foreign key setting that will break)
- Currently, this code will not add foreign keys when creating tables. A second call to autoupdate is required with the current code. I see a couple possible ways to fix this.
  1. Run `SET FOREIGN_KEY_CHECKS=0;` at the beginning of automigrate and `SET FOREIGN_KEY_CHECKS=1;` at the end.
  2. Enforce some sort of order to the table creation so that a model is not run through create until after its dependencies have been created (this sounds difficult).
  3. Automatically re-run the foreign key code at the end of automigrate/autoupdate if there are any new tables
- Should there be a setting to not create foreign keys even if relations are being used?

NOTE: I did add a new set of tests to validate that the foreign keys build and drop correctly. However, per the issue above, I have to back-to-back call automigrate and autoupdate to get the foreign keys to build for the new model.

Let me know what more I can do, what I should change, which direction should be pursued for enabling foreign keys on table creation, etc.

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -493,8 +493,6 @@ function mixinMigration(MySQL, mysql) {
   };
 
   MySQL.prototype.buildForeignKeyDefinition = function(model, keyName) {
-    //TODO validate that these are in the same DB, etc.
-    //TODO this is probably a better place to check this than in autoupdate or buildColumnDefinitions
     var definition = this.getModelDefinition(model);
 
     var fk = definition.settings.foreignKeys[keyName];
@@ -551,13 +549,6 @@ function mixinMigration(MySQL, mysql) {
         sql.push(i);
       });
 
-//       var foreignKeys = definition.settings.foreignKeys || {};
-//       for (var fk in foreignKeys) {
-//         var constraint = self.buildForeignKeyDefinition(model, fk);
-//         if (constraint) {
-//           sql.push(constraint);
-//         }
-//       }
       return sql.join(',\n  ');
     };
 
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -418,15 +418,19 @@ function mixinMigration(MySQL, mysql) {
     //TODO validate that these are in the same DB, etc.
     //TODO this is probably a better place to check this than in autoupdate or buildColumnDefinitions
     var definition = this.getModelDefinition(model);
+
     var fk = definition.settings.foreignKeys[keyName];
     if (fk) {
-      var fkEntity = typeof fk.entity === 'object' ? fk.entity : {name: fk.entity}; //this.getModelDefinition(fk.entity);
+      //get the definition of the referenced object
+      var fkEntityName = (typeof fk.entity === 'object') ? fk.entity.name : fk.entity;
 
-      //TODO verify that the other model in the same DB
-      return ' CONSTRAINT ' + this.client.escapeId(fk.name) +
-        ' FOREIGN KEY (' + fk.foreignKey + ')' +
-        ' REFERENCES ' + this.tableEscaped(fkEntity.name) +
-        '(' + this.client.escapeId(fk.entityKey) + ')';
+      //verify that the other model in the same DB
+      if (this._models[fkEntityName]) {
+        return ' CONSTRAINT ' + this.client.escapeId(fk.name) +
+          ' FOREIGN KEY (' + fk.foreignKey + ')' +
+          ' REFERENCES ' + this.tableEscaped(fkEntityName) +
+          '(' + this.client.escapeId(fk.entityKey) + ')';
+      }
     }
     return '';
   };
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -42,11 +42,13 @@ function mixinMigration(MySQL, mysql) {
       var table = self.tableEscaped(model);
       self.execute('SHOW FIELDS FROM ' + table, function(err, fields) {
         self.execute('SHOW INDEXES FROM ' + table, function(err, indexes) {
-          if (!err && fields && fields.length) {
-            self.alterTable(model, fields, indexes, done);
-          } else {
-            self.createTable(model, done);
-          }
+          self.discoverForeignKeys(model, {}, function(discoverErr, foreignKeys) {
+            if (!err && fields && fields.length) {
+              self.alterTable(model, fields, indexes, foreignKeys, done);
+            } else {
+              self.createTable(model, done);
+            }
+          });
         });
       });
     }, cb);
@@ -93,14 +95,16 @@ function mixinMigration(MySQL, mysql) {
       var table = self.tableEscaped(model);
       self.execute('SHOW FIELDS FROM ' + table, function(err, fields) {
         self.execute('SHOW INDEXES FROM ' + table, function(err, indexes) {
-          self.alterTable(model, fields, indexes, function(err, needAlter) {
-            if (err) {
-              return done(err);
-            } else {
-              ok = ok || needAlter;
-              done(err);
-            }
-          }, true);
+          self.discoverForeignKeys(model, {}, function(discoverErr, foreignKeys) {
+            self.alterTable(model, fields, indexes, foreignKeys, function(err, needAlter) {
+              if (err) {
+                return done(err);
+              } else {
+                ok = ok || needAlter;
+                done(err);
+              }
+            }, true);
+          });
         });
       });
     }, function(err) {
@@ -111,7 +115,7 @@ function mixinMigration(MySQL, mysql) {
     });
   };
 
-  MySQL.prototype.alterTable = function(model, actualFields, actualIndexes, done, checkOnly) {
+  MySQL.prototype.alterTable = function(model, actualFields, actualIndexes, actualFks, done, checkOnly) {
     var self = this;
     var m = this.getModelDefinition(model);
     var propNames = Object.keys(m.properties).filter(function(name) {
@@ -120,6 +124,10 @@ function mixinMigration(MySQL, mysql) {
     var indexes = m.settings.indexes || {};
     var indexNames = Object.keys(indexes).filter(function(name) {
       return !!m.settings.indexes[name];
+    });
+    var newFks = m.settings.foreignKeys || {};
+    var newFkNames = Object.keys(newFks).filter(function(name) {
+      return !!m.settings.foreignKeys[name];
     });
     var sql = [];
     var ai = {};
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

superkhau:
Can you add some tests to verify your changes and prevent regressions in the future?

matthewdickinson:
Oops. I forgot to commit/push that change. I just pushed it.

superkhau:
@matthewdickinson Is that the test that verifies your PR? I just see a TODO comment.

matthewdickinson:
@superkhau Nope. I was just stupid an committed the wrong file. I just pushed the correct one now.

superkhau:
test please

matthewdickinson:
@superkhau The tests (`make test`) still pass locally. How do I fix the test here? I see that my commit messages are a problem in the commit linter; what's the best way to fix that?

superkhau:
@matthewdickinson Sorry for some of the confusion, we have to remove the Makefile altogether. You should be running tests via `npm test` (which in turn also runs the linter -- see `package.json`).

# Commit linter

See https://github.com/strongloop/loopback-contributor-docs/blob/master/git-commit-messages.md#git-commit-message-guidelines

```
**
**  Linting commit logs
**
**  4 problems found:
**    406af99 - Added test to verify that foreign key creation work: First line should be 50 characters or less (saw 64)
**    356ed9d - Updated automatic foreign key creation to correctly: First line should be 50 characters or less (saw 95)
**    3762625 - Updated alterTable to actually run the drop constra: First line should be 50 characters or less (saw 60)
**    a183dc4 - Added rough handling for auto-creating/dropping for: First line should be 50 characters or less (saw 85)
**
```

# cis-jenkins

These are unrelated to your changes, so it's safe to ignore these for now.

matthewdickinson:
@superkhau What is the best way to fix the commit messages? Should I just amend the commits and force push (will that screw up the pull request), should I create a new pull request?

Thanks.

superkhau:
>@superkhau What is the best way to fix the commit messages? Should I just amend the commits and force push (will that screw up the pull request), should I create a new pull request?

Yes, you can amend the commit and force push to overwrite all changes in this PR to what your local looks like. Do not create a new PR.

matthewdickinson:
OK. Will do

matthewdickinson:
I've fixed the one issue I listed above (this will now not allow references to entities in other connectors).

@superkhau  The remaining issue is trying to determine how to handle foreign keys when creating a table. When creating multiple tables, it's possible that there will be a foreign key in the current table that references a table that hasn't been created yet. How should that be handled?
Currently, this code will not add foreign keys when creating tables. A second call to autoupdate is required with the current code. I see a couple possible ways to fix this.

- Run SET FOREIGN_KEY_CHECKS=0; at the beginning of automigrate and SET FOREIGN_KEY_CHECKS=1; at the end.
- Enforce some sort of order to the table creation so that a model is not run through create until after its dependencies have been created (this sounds difficult).
- Automatically re-run the foreign key code at the end of automigrate/autoupdate if there are any new tables

superkhau:
This is an interesting problem. I would say the first choice is the simplest but does not enforce referencial integrity (meaning your tables may not actually work after migration). Number 2 is the *best* solution, but as you said, difficult to ensure proper ordering. I think we need a way to run the migration, enforce referencial integrity and roll back on failure if any foreign key checks fail or at least warn users so they can manually fix it.

@raymondfeng @bajtos @strongloop/fa-db-connectors Thoughts?

matthewdickinson:
The other options that I could see if to separate out the foreign key _creation_ so that it happened at the end of the create/migrate loop after all of the tables/indexes have been created. That would obviously require a bit of restructuring of this implementation, but I think it would also solve the issue.

lib/migration.js
```diff
@@ -42,11 +42,13 @@ function mixinMigration(MySQL, mysql) {
       var table = self.tableEscaped(model);
       self.execute('SHOW FIELDS FROM ' + table, function(err, fields) {
         self.execute('SHOW INDEXES FROM ' + table, function(err, indexes) {
-          if (!err && fields && fields.length) {
-            self.alterTable(model, fields, indexes, done);
-          } else {
-            self.createTable(model, done);
-          }
+          self.discoverForeignKeys(self.table(model), {}, function(discoverErr, foreignKeys) {
```
bajtos: unhandled (ignored) `discoverErr`

lib/migration.js
```diff
@@ -93,14 +95,16 @@ function mixinMigration(MySQL, mysql) {
...
         self.execute('SHOW INDEXES FROM ' + table, function(err, indexes) {
-          self.alterTable(model, fields, indexes, function(err, needAlter) {
-            if (err) {
-              return done(err);
-            } else {
-              ok = ok || needAlter;
-              done(err);
-            }
-          }, true);
+          self.discoverForeignKeys(self.table(model), {}, function(discoverErr, foreignKeys) {
```
bajtos: unhandled (ignored) `discoverErr`

lib/migration.js
```diff
@@ -111,7 +115,7 @@ function mixinMigration(MySQL, mysql) {
     });
   };
 
-  MySQL.prototype.alterTable = function(model, actualFields, actualIndexes, done, checkOnly) {
+  MySQL.prototype.alterTable = function(model, actualFields, actualIndexes, actualFks, done, checkOnly) {
```
bajtos: Since this is changing (semi)public API exposed by the mysql connector, should we keep supporting the old method signature `alterTable(model, actualFields, actualIndexes, done, checkOnly)` for backwards compatibility?

test/mysql.autoupdate.test.js
```diff
@@ -167,6 +167,152 @@ describe('MySQL connector', function() {
...
+                if (err) return done (err);
+
+                assert.equal(props.length, 3);
+
+                ds.connector.execute(foreignKeySelect, function(err, updatedForeignKeys) {
+                  if (err) return done (err);
+                  assert(updatedForeignKeys);
+                  assert(updatedForeignKeys.length.should.be.equal(0));
+
+                  done(err, result);
```
bajtos: We usually call `done()` without any arguments when the test passed.

bajtos:
I did a quick review of the change, my knowledge of this code base is not good enough to be able to provide a meaningful feedback.

At high level:

Does this patch support `autoupdate` too, where the database is updated while existing data is preserved?

How expensive it is to rebuild a foreign key in MySQL? I assume MySQL needs to maintain some sort of an index for each foreign key. AFAIK, (re)building a new index from scratch is pretty expensive operation. As a result, the current implementation that drops & rebuilds foreign keys may be very slow.

As for enforcing referencial integrity, I like these two proposal mentioned above:
 1. create/update the tables in such order that foreign keys can be created/updated at the same time
 2. move the creation/updating of foreign keys to happen only after all tables were already created/updated

matthewdickinson:
Yes, the patch is designed to support autoupdate as well. Also, the foreign keys should not be dropped and re-created unless there is a change in the relation. Otherwise, it will leave them untouched.

test/mysql.autoupdate.test.js
matthewdickinson: I was just trying to match the convention in the rest of that file. I can make that change.

lib/migration.js
matthewdickinson: What's the best/preferred way to do that? Should it just do typeof checks for the last 3 parameters to see which is which?

lib/migration.js
bajtos: > How should a foreign key discovery error be handled here?

Honestly, I don't know, because I am not familiar with this code base. Is it a bad idea to abort the whole autoupdate/automigrate operation in such case? If not, then a warning/error should be printed to the console at least.

test/mysql.autoupdate.test.js
bajtos: Ah, ok, in that case it may be better to stay consistent. I'll leave the judgment to you and @superkhau

bajtos:
@slnode ok to test

lib/migration.js
matthewdickinson: OK. Does the new code look appropriate?

matthewdickinson:
@bajtos  @superkhau OK. I've moved the code for adding foreign keys so that it runs after all of the tables have been created (this should avoid broken foreign keys). However, at the moment, `automigrate` does not support adding foreign keys; this is because the function is declared in `loopback-connector/lib/sql.js`. I'm assuming that we could overwrite this function in the `mysql-connector`, but is that the functionality desired?

Also, I see that some of the checks are now failing, but I'm not sure how to debug these. The tests in `npm test` all pass; is there a way to turn on these tests as well, or are these completely different?

matthewdickinson:
@bajtos @superkhau Is there anything that I can do to fix these failing test? As I said, they don't seem to run on my machine (everything passes); is there a way to enable them?

slnode:

superkhau:
@matthewdickinson Sorry I haven't been able to look into this PR more for awhile -- I will take a look at the new updates soon. Regarding CI, it won't go green because master is failling OOB ATM -- @loay has been working on it for awhile now I think. @loay Any insight into when the MySQL will be green?

loay:
@slnode ok to test

loay:
@slnode ok to test

loay:
@matthewdickinson @superkhau 
Just  run the tests again. Everything passes now. 
PR needs to be rebased and I will give the code another review as well.

matthewdickinson:
@superkhau @loay I've rebased the branch.

ssh24:
@matthewdickinson Can you please rebase your branch?

matthewdickinson:
@ssh24 I just rebased it.

ssh24:
@matthewdickinson Thank you. I am going to do the code review soon. So far, looks good.

lib/migration.js
```diff
@@ -348,6 +486,33 @@ function mixinMigration(MySQL, mysql) {
...
   };
 
+  MySQL.prototype.getAlterStatement = function(model, statements) {
+    return statements.length ?
+      'ALTER TABLE ' + this.tableEscaped(model) + ' ' + statements.join(',\n') :
+      '';
+  };
+
+  MySQL.prototype.buildForeignKeyDefinition = function(model, keyName) {
+    //TODO validate that these are in the same DB, etc.
```
matthewdickinson: Sorry. It's actually was already about 6 lines down from that. I just removed the TODO

lib/migration.js
```diff
@@ -385,6 +550,14 @@ function mixinMigration(MySQL, mysql) {
       indexes.forEach(function(i) {
         sql.push(i);
       });
+
+//       var foreignKeys = definition.settings.foreignKeys || {};
```
matthewdickinson: @ssh24 Done

slnode:

slnode:

slnode:

slnode:

PR #194, opened 2016-09-20
Remove Makefile in favour of NPM scripts

superkhau:
We should use `npm test` over the deprecated Makefile test targets.

Also prevent some confusion at https://github.com/strongloop/loopback-connector-mysql/pull/193#issuecomment-248333766

@strongloop/fa-db-connectors PTAL


superkhau:
Failure is unrelated to this PR. @loay Did you get to pull in this issue into this sprint?

PR #195, opened 2016-09-22
Update deps to loopback 3.0.0 RC

bajtos:
Connect to strongloop-internal/scrum-loopback#1063


bajtos:
AFAICT, the test failures are not related to my change, as they are present in older CI builds too. I am going to land the patch despite the red CI.

PR #202, opened 2016-10-05
follow mysql recommendations for handling booleans

azatoth:
Per http://dev.mysql.com/doc/refman/5.7/en/numeric-type-overview.html
`BOOL` and `BOOLEAN` is alias for `TINYINT(1)` thus we should make
sure such are handled as such.
There is no indication that `CHAR(1)` is used for boolean values, and
can more often interfer with usage to store a single character in the
database ("n", etc...); Thus we remove that exception.
Also as `BIT(1)` practically is a boolean value, we default to use it as
such and mark it as `Boolean`.


slnode:

slnode:

superkhau:
Can you add some tests to verify your changes and prevent regressions in the future?

azatoth:
@superkhau I did add some tests; are they not sufficient?

lib/discovery.js
```diff
@@ -406,18 +408,12 @@ function mixinDiscovery(MySQL, mysql) {
...
     var type = mysqlType.toUpperCase();
     switch (type) {
       case 'CHAR':
-        if (dataLength === 1) {
-          // Treat char(1) as boolean
-          return 'Boolean';
-        } else {
-          return 'String';
-        }
-        break;
```
superkhau: Why are we removing this?

superkhau:
@azatoth Yes you did, sorry I didn't notice earlier. ;) The changes are mostly LGTM, but I would like to know why you removed the `char(1)` check before landing.

lib/discovery.js
azatoth: First of I couldn't find any recommendation to use CHAR(1) as a data type for a boolean. Also I've found many situations where CHAR(1) is used to store a single letter (e.g. "m", "c", etc...) or as an old style bit array. Converting it to a boolean I believe is not the most common usecase.

raymondfeng: I think folks use `Y`, 'N', 'T', 'F', '0', '1' for boolean too. We should keep it for backward compatibility.

azatoth: @raymondfeng if the current system interpret these characters as correct boolean values, then I can understand that it is reasonable to have it mapped as a boolean; Though I wonder if it would be possible to add a switch to change the behavior?

azatoth:
@raymondfeng in e54840db0788464b69f4a48445f142d976ea5363 I've added a options flag `treatCHAR1AsBoolean` to allow the usecase with character representing booleans to be handled.
I could have reversed it as `treatCHAR1AsString` but I felt calling it out as Boolean was more explicit.

superkhau:
@azatoth I think this is changing the old behaviour and thus is a breaking change -- Can we default to the old behaviour and use a flag to allow users to switch to your new behaviour instead? Besides that the feature flag you're adding LGTM.

azatoth:
@superkhau That is possible, I just felt it would sound strange with a variable called "treatCHAR1AsString" but any direction works;
I assume it's logical then to add "treatBIT1AsBit" and "treatTINYINT1AsTinyInt" as well then? or should they be the other way and have different logic?

superkhau:
- treatCHAR1AsString: true should be default
- treatBIT1AsBit: false should be default
- treatTINYINT1AsTinyInt: false should be default

This preserves all the old behaviour. We should then add a deprecation notice for LoopBack 4 to drop these and default to the recommended behaviours you're proposing.

Another possibility is if we can think of a way to on ALL 3 via a single flag -- `useMySQL57recommendedBooleanScheme: true` and set the logic to act accordingly when on and preserve old behaviour when off.

@bajtos @raymondfeng Thoughts?

superkhau:
I think Booleans also defaulted to `String` before too

lib/discovery.js
bajtos: Backwards compatibility is important, please preserve it.

> I wonder if it would be possible to add a switch to change the behavior?

I am fine with that, this is the approach we are usually taking - add a configuration option allowing users to opt into backwards-incompatible behaviour.

bajtos:
> - treatCHAR1AsString: true should be default
- treatBIT1AsBit: false should be default
- treatTINYINT1AsTinyInt: false should be default

> This preserves all the old behaviour. We should then add a deprecation notice for the next major to drop these and default to the recommended behaviours you're proposing.

> Another possibility is if we can think of a way to on ALL 3 via a single flag -- `useMySQL57RecommendedBooleanScheme: true` (can't think of a better name) and set the logic to act accordingly when on and preserve old behaviour when off.

@superkhau I don't have a strong opinion. As long as we preserve backwards compatibility, I am fine with whatever settings make sense to you and other @strongloop/fa-db-connectors members.

superkhau:
@strongloop/fa-db-connectors Ideas for flag name? Opinions on https://github.com/strongloop/loopback-connector-mysql/pull/202#issuecomment-253662753?

test/mysql.discover.test.js
```diff
@@ -271,4 +277,34 @@ describe('Discover and build models', function() {
...
+      var schema = models.Inventory.definition;
+      assert(schema.properties.enabled);
+      assert.strictEqual(schema.properties.enabled.type, Boolean);
+    });
+
+    it('should be able to find an instance', function(done) {
+      assert(models.Inventory, 'Inventory model must exist');
+      models.Inventory.findOne(function(err, inv) {
+        assert(!err, 'error should not be reported');
+        done();
```
Amir-61: I would change:
```
        assert(!err, 'error should not be reported');
        done();
```
to:
```
        done(err);
```

test/mysql.discover.test.js
```diff
@@ -271,4 +277,34 @@ describe('Discover and build models', function() {
...
+        visited: {},
+        associations: true,
+        treatCHAR1AsBoolean: true,
+      }, function(err, models_) {
+        models = models_;
+        done(err);
+      });
+    });
+
+    it('should discover and build models', function() {
```
superkhau: it('discovers and builds models'

test/mysql.discover.test.js
```diff
@@ -271,4 +277,34 @@ describe('Discover and build models', function() {
...
+      assert(models.Inventory, 'Inventory model should be discovered and built');
+      var schema = models.Inventory.definition;
+      assert(schema.properties.enabled);
+      assert.strictEqual(schema.properties.enabled.type, Boolean);
+    });
+
+    it('should be able to find an instance', function(done) {
```
superkhau: it('finds instances

test/mysql.discover.test.js
superkhau: actually, we can reduce more:

```
models.Inventory.findOne(done);

**Comment by Amir-61** [comment]
I'm fine with the names you came up with in https://github.com/strongloop/loopback-connector-mysql/pull/202#issuecomment-253662753; they are explicit enough.

**Comment by Amir-61** [comment]
@azatoth your PR needs to be rebased on top of master and some merge conflicts need to be fixed as well.

**Comment by Amir-61** [comment]
@superkhau I added a few comments on the PR; You can take it from here...

File: test/mysql.discover.test.js (hunk 3 above)
Comment (azatoth) [inline test/mysql.discover.test.js]: @Amir-61 I used 'should' as all other tests in test/mysql.discover.test.js uses it("should...");

**Comment by Amir-61** [comment]
@superkhau I believe this needs to be documented after landing as well; I'm going to addd `needs doc` label for this PR as well...

CC: @crandmck

**Comment by superkhau** [comment]
@azatoth I think the changes are almost good to land with my final round of comments. Thanks for your patience.

**Comment by crandmck** [comment]
Should be documented here: http://loopback.io/doc/en/lb2/MySQL-connector.html#type-mappings
@Amir-61 would you be able to update the docs please?

File: lib/discovery.js [hunk 5]
```diff
@@ -240,6 +242,19 @@ function mixinDiscovery(MySQL, mysql) {
     }
     table = args.table;
     options = args.options;
+
+    var default_options = {
```
azatoth: Seems the local .eslintrc has a `{"camelcase": 0}` rule...

test/mysql.discover.test.js
```diff
@@ -271,4 +271,129 @@ describe('Discover and build models', function() {
       done();
     });
   });
+
+  describe('discoverModelProperties() flags', function() {
+    describe('with default flags', function() {
```
azatoth: as `context` is just an alias for `describe`, is it necessary?

lib/discovery.js
superkhau: Feel free to leave it, the rule is wrong and shouldn't be there. I will submit another PR to fix later (I don't want you to have to rebase again if I submit the fix now).

test/mysql.discover.test.js
superkhau: It's definitely subjective, but we documented the convention we decided to go with at https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#good-12

Amir-61:
@crandmck 

>Should be documented here: http://loopback.io/doc/en/lb2/MySQL-connector.html#type-mappings
>@Amir-61 would you be able to update the docs please?

Sure I'll document it after it gets landed; just in case not to forget, I created an issue to remember: https://github.com/strongloop/loopback-connector-mysql/issues/208

azatoth:
@superkhau Do you want me to rebase the changes into the main commit or do you want it appended?

superkhau:
@azatoth Rebase and squash please.

superkhau:
test please
@slnode test please

superkhau:
@slnode test please

slnode:

test/mysql.discover.test.js
```diff
@@ -271,4 +271,137 @@ describe('Discover and build models', function() {
...
+        assert.strictEqual(schema.properties.active.type, Boolean);
+      });
+
+      function discoverAndBuildModels(done) {
+        db.discoverAndBuildModels('INVENTORY', {
+          owner: 'STRONGLOOP',
+          visited: {},
+          associations: true,
+          treatTINYINT1AsTinyInt: false,
+        }, function(err, models_) {
```
superkhau:
```
if (err) return done(err);
models = models_;
schema = models.Inventory.definition;
done();
```
but I'll help fix this

superkhau:
Helped land PR at https://github.com/strongloop/loopback-connector-mysql/pull/232

bajtos:
> @superkhau Helped land this PR at #232 

FYI, since September 2016, one can modify the feature branch used to create a pull request, thus it's no longer necessary to create new pull requests. See this announcement post for more details:
https://github.com/blog/2247-improving-collaboration-with-forks

Issue #217, opened 2016-12-02
JSON parser issue


chinuy:
at https://github.com/strongloop/loopback-connector-mysql/blob/master/lib/mysql.js#L435
No try/catch block in JSON.parse(val). 

I wonder is there any reason not to do so? I got an issue that my db data is corrupt and therefore, the parse will raise exception and crash system every time. I would more like the return is `null` if the JSON.parse() function fail so that I still have chance to just overwrite data.

superkhau:
No reason, probably a oversight -- would you like to submit a PR?

chinuy:
#218 PR created.

dhmlau:
PR #218 is closed with the last comment on Dec 8, 2016: 
`@chinuy Let's close this PR as rejected and you can submit a new PR with the suggested fixed. Feel free to ping @bajtos or me directly when you need another review.`

@chinuy , would you like to submit another PR with the suggestion on #218?  Thanks.

chinuy:
Not in this repo. It would be [loopback-connector](https://github.com/strongloop/loopback-connector). Per discussion on #218, I was stuck in how to handle [`filter.include`](https://github.com/strongloop/loopback-connector-mysql/pull/218#issuecomment-265880248).

PR #218, opened 2016-12-03
issue #217, JSON parser with try/catch

chinuy:
Turns out the testing is a little bit tricky. Please verify before merge. Thanks.

### DIFF
--- a/test/mysql.test.js
+++ b/test/mysql.test.js
@@ -81,6 +81,23 @@ describe('mysql', function() {
     });
   });
 
+  it('should return null if JSON.parse fail', function(done) {
+    Post.create({title: 'a', history: {a: 1, b: 'b'}}, function(err, post) {
+      should.not.exist(err);
+      
+      // break the content
+      var sql = 'UPDATE ' + global.getConfig().database + '.PostWithDefaultId SET history = \'{corrupted,\' where title=\'a\'';
+      db.connector.query(sql, function(err, res) {
+        Post.findById(post.id, function(err, p) {
+          p.id.should.be.equal(post.id);
+          should(p.history).be.null;
+
+          done();
+        });
+      });
+    });
+  });
+
   it('should allow ObjectID', function(done) {
     var uid = new ObjectID('123');
     Post.create({title: 'a', content: 'AAA', userId: uid},
--- a/test/mysql.test.js
+++ b/test/mysql.test.js
@@ -81,17 +81,17 @@ describe('mysql', function() {
     });
   });
 
-  it('should return null if JSON.parse fail', function(done) {
+  it('return null if JSON parsing fails', function(done) {
     Post.create({title: 'a', history: {a: 1, b: 'b'}}, function(err, post) {
+      if (err) return done(err);
       should.not.exist(err);
       
       // break the content
-      var sql = 'UPDATE ' + global.getConfig().database + '.PostWithDefaultId SET history = \'{corrupted,\' where title=\'a\'';
-      db.connector.query(sql, function(err, res) {
-        Post.findById(post.id, function(err, p) {
-          p.id.should.be.equal(post.id);
-          should(p.history).be.null;
-
+      var statement = 'UPDATE ' + global.getConfig().database + '.PostWithDefaultId SET history = \'{corrupted,\' where title=\'a\'';
+      db.connector.query(statement, function(err, res) {
+        Post.findById(post.id, function(err, post) {
+          if (err) return done(err);
+          should.not.exist(post.history);
           done();
         });
       });
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -433,7 +433,11 @@ MySQL.prototype.fromColumnValue = function(prop, val) {
       case 'Object':
       case 'JSON':
         if (typeof val === 'string') {
-          val = JSON.parse(val);
+          try {
+            val = JSON.parse(val);
+          } catch(err) {
+            val = null;
+          }
         }
         break;
       default:


slnode:

slnode:

slnode:

slnode:

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
     });
   });
 
+  it('should return null if JSON.parse fail', function(done) {
+    Post.create({title: 'a', history: {a: 1, b: 'b'}}, function(err, post) {
+      should.not.exist(err);
```
superkhau: `if (err) return done(err);` for better stack traces

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
     });
   });
 
+  it('should return null if JSON.parse fail', function(done) {
```
superkhau: it('returns null if JSON parsing fails'...

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
...
+    Post.create({title: 'a', history: {a: 1, b: 'b'}}, function(err, post) {
+      should.not.exist(err);
+      
+      // break the content
+      var sql = 'UPDATE ' + global.getConfig().database + '.PostWithDefaultId SET history = \'{corrupted,\' where title=\'a\'';
```
superkhau: var statement = ...

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
...
+      db.connector.query(sql, function(err, res) {
```
superkhau: sql -> statement

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
...
+        Post.findById(post.id, function(err, p) {
+          p.id.should.be.equal(post.id);
+          should(p.history).be.null;
+
```
superkhau: remove empty space

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
...
   });
 
+  it('should return null if JSON.parse fail', function(done) {
+    Post.create({title: 'a', history: {a: 1, b: 'b'}}, function(err, post) {
+      should.not.exist(err);
+      
+      // break the content
+      var sql = 'UPDATE ' + global.getConfig().database + '.PostWithDefaultId SET history = \'{corrupted,\' where title=\'a\'';
+      db.connector.query(sql, function(err, res) {
+        Post.findById(post.id, function(err, p) {
```
superkhau: if (err) return done(err);
p -> post

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
...
+          p.id.should.be.equal(post.id);
+          should(p.history).be.null;
```
superkhau: should.not.exist(post.history);

chinuy:
test case enhanced with @superkhau's comments

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
     });
   });
 
+  it('return null if JSON parsing fails', function(done) {
```
superkhau: plural -> returns

test/mysql.test.js
```diff
@@ -81,6 +81,23 @@ describe('mysql', function() {
...
+    Post.create({title: 'a', history: {a: 1, b: 'b'}}, function(err, post) {
+      if (err) return done(err);
+      should.not.exist(err);
```
superkhau: remove should.not.exist(err); line as if (err) return done(err) does it all.

chinuy:
Do you ask for returning error in the `lib/mysql.js`?
Assigning `null` is only because the `MySQL.prototype.fromColumnValue()` function needs to return a `val`.  Current try/catch mechanism only prevents it from crashing main program.

superkhau:
@chinuy Depends how you see the original code -- If we want to keep things sync, we should be try/catching from the caller(s) and not in the code via coercing to null. In this case throwing is already correct. If we were to treat it async, we would return 'err' normally via callback. Those are the two choices I see, what are your thoughts?

chinuy:
@superkhau  Agree. This can be many use cases, depends on users. I believe try/catch is necessary. Just where to put it, and what kind of message should deliver to the user.

A more complicated case might be:
a Model.find() function is call, with just one row containing corrupted data. What's the best thing to return? bare in mind that any error will block all following logic, which isn't always good for a production system.

To me, as an user, I hope the system:

1. return a `null` to me so my system can keep running
2. log such event in my log file so I can inspect the problem in the future

chinuy:
I want to withdraw a little. Maybe throwing out an error to user is a good way, as long as the user can distinguish whether err is critical. Your thought?

superkhau:
My thought is to refactor this code to support callback style so we can return any error to using expected nodejs patterns. function(err, ...) so the user can handle in the callback or log to some file in your use case.

@bajtos Thoughts?

bajtos:
I think there are two parts of this problem:

(1) How to signal the problem to the user? The first solution that comes to my mind is that `Post.findById` in the newly added tests should fail with an error. This is follows the Node.js pattern - if there is something wrong, then fail with an error.

This has a caveat though: the code calling `findById` cannot easily fix the problem, i.e. the app user will keep seeing 500 Internal Error responses when querying the malformed record. The current proposal, where we convert invalid field to `null`, may be better in this light. But then it hides the problem, which is not great either.

(2) Assuming we decide that `findById` should return an error when a JSON value cannot be parsed: I think converting `fromColumnValue` to callback-style is overkill and it will only complicate the code base. IMO, we should move the try-catch block to a different place (possibly to multiple places - everywhere we call `fromColumnValue`).

AFAICT, `fromColumnValue` is used by `fromRow` ([source](https://github.com/strongloop/loopback-connector/blob/c6a68ec5d8c94f293f04eab6986e69113797ebc3/lib/sql.js#L1168)) which is in turn used by `all` ([source](https://github.com/strongloop/loopback-connector/blob/c6a68ec5d8c94f293f04eab6986e69113797ebc3/lib/sql.js#L1200)).

I think this issue affects other database connectors beyond MySQL, possibly all SQL connectors, therefore it's a good idea to fix the issue in the shared connector base (see the links I posted).

superkhau:
>IMO, we should move the try-catch block to a different place (possibly to multiple places - >everywhere we call fromColumnValue).
>
>AFAICT, fromColumnValue is used by fromRow (source) which is in turn used by all (source).
>
>I think this issue affects other database connectors beyond MySQL, possibly all SQL connectors, >therefore it's a good idea to fix the issue in the shared connector base (see the links I posted).

I also think this is the correct solution. Keep the code sync and let the callers handle the try/catches and return errors accordingly. Which is pain to refac because it touches all the places you mentioned. ;( @chinuy Thats the solution. ;) Would you like to try?

bajtos:
> Which is pain to refac because it touches all the places you mentioned. ;( 

Not really. I believe we only need to fix [SQLConnector.prototype.all](https://github.com/strongloop/loopback-connector/blob/c6a68ec5d8c94f293f04eab6986e69113797ebc3/lib/sql.js#L1200):

```js
try {
  var objs = data.map(function(obj) {
    return self.fromRow(model, obj)
  });
} catch (err) {
  return cb(err);
}
```

chinuy:
how about using `async`, so rows with correct data can be preserved, while rows with malformed data will return null

such as:
```js
async.map(objs, function(obj, done) {
  try {
    done(null, self.fromRow(model, obj));
  } catch(err) {
    done(err, null);
  }
}, function(err, res) {

});
```

superkhau:
@chinuy I agree with @bajtos's comment at https://github.com/strongloop/loopback-connector-mysql/pull/218#issuecomment-265697182 to modify only `prototype.all` and return the error to the user (ie. for you to handle/change to null there).

chinuy:
https://github.com/chinuy/loopback-connector/commit/678500c2f24f81d5ab8fef9cac44566d60a25b31

I try to fix this problem based on @bajtos 's comment. However, the callback involved `filter.include` seems tangled.

superkhau:
@chinuy Let's close this PR as rejected and you can submit a new PR with the suggested fixed. Feel free to ping @bajtos or me directly when you need another review.

PR #229, opened 2017-01-10
Update readme w info from docs

crandmck:
Per https://github.com/strongloop/loopback.io/issues/156 I am moving content of the doc into the README, where we'll use a script to pull it back into the doc.  Then we only have to maintain it in one place, here.

@0candy

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,50 +1,324 @@
 # loopback-connector-mysql
 
-`loopback-connector-mysql` is the MySQL connector module for [loopback-datasource-juggler](https://github.com/strongloop/loopback-datasource-juggler/).
+[MySQL](https://www.mysql.com/) is a popular open-source relational database management system (RDBMS).  The `loopback-connector-mysql` module provides the MySQL connector module for the LoopBack framework.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,50 +1,324 @@
 # loopback-connector-mysql
 
-`loopback-connector-mysql` is the MySQL connector module for [loopback-datasource-juggler](https://github.com/strongloop/loopback-datasource-juggler/).
+[MySQL](https://www.mysql.com/) is a popular open-source relational database management system (RDBMS).  The `loopback-connector-mysql` module provides the MySQL connector module for the LoopBack framework.
 
-For complete documentation, see [StrongLoop Documentation | MySQL Connector](http://loopback.io/doc/en/lb2/MySQL-connector.html).
+<div class="gh-only">See also [LoopBack MySQL Connector](http://loopback.io/doc/en/lb2/MySQL-connector.html) in LoopBack documentation.
```
0candy: Should we use lb3 link? http://loopback.io/doc/en/lb3/MySQL-connector.html

crandmck:
Good catch on the link!  I also improved the "discovery and automigration" section, which is mostly links to the docs (and will be largely the same across the supporting connectors)

PR #241, opened 2017-02-13
Refactor Discovery Methods

loay:
### Description
Refactor discovery test cases

#### Related issues
connect to https://github.com/strongloop/loopback-connector/issues/15

### DIFF
--- a/lib/discovery.js
+++ b/lib/discovery.js
@@ -44,19 +44,19 @@ function mixinDiscovery(MySQL, mysql) {
    * @params {Object} [options] Options object
    * @returns {String} The SQL statement
    */
-  function querySchemas(options) {
+  MySQL.prototype.buildQuerySchemas = function(options) {
     var sql = 'SELECT catalog_name as "catalog",' +
       ' schema_name as "schema"' +
       ' FROM information_schema.schemata';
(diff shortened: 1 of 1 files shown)


loay:
@slnode test please

loay:
@slnode test please

loay:
@slnode test please

loay:
@slnode test please

lib/discovery.js
```diff
@@ -130,44 +121,14 @@ function mixinDiscovery(MySQL, mysql) {
...
-  };
 
   /*!
    * Normalize the arguments
    * @param table string, required
    * @param options object, optional
    * @param cb function, optional
    */
-  function getArgs(table, options, cb) {
+  MySQL.prototype.getArgs = function(table, options, cb) {
```
raymondfeng: Shouldn't we push this to the base SQL connector? Do you see the need to have connector-specific implementations?

loay: Unfortunately, we cannot do that for now, as each connector has its implementation and it will require more digging. we took note as to work on it in version 2.0 of the refactoring. 
for example, db2 allows `table` to be empty instead of giving error: https://github.com/strongloop/loopback-connector-db2/blob/master/lib/discovery.js#L191

PR #248, opened 2017-03-04
Add checkFieldAndIndex for table status

ssh24:
### Description

Related code: https://github.com/strongloop/loopback-connector/blob/bba29f88f2a5cd23ae31d266c46f2a026503619e/lib/sql.js#L236
Each sql connector has its own check of `fields` and `indexes` exist method, then determine callback the result or not.


jannyHou:
@slnode test please

ssh24:
Closing. Unnecessary PR.

PR #251, opened 2017-03-15
Refactor alter table

ssh24:
This PR refactors the `alterTable` function in a way to unify the logic between all connectors. Currently, the `alterTable` function is about __300__ lines long in mysql. This refactor extracts the code and puts them into smaller functions like the postgresql does.

**Note:** Mysql supports auto-update/migrate with foreign keys, hence the unification of the `alterTable` function includes foreign key functions.

List of functions that were extracted out to be put as a smaller function and called in `alterTable`:
 - **getAddModifyColumns**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L159
 - **getColumnsToAdd**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L166
 - **getDropColumns**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L223
 - **getColumnsToDrop**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L230
 - **addIndexes**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L254
 - **applySqlChanges**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L409
 - **getForeignKeySQL**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L418
 - **addForeignKeys**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L436
 - **dropForeignKeys**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L459
 - **alterTable**- https://github.com/strongloop/loopback-connector-mysql/blob/refactor-alter-table/lib/migration.js#L505

After breaking down the logic in `alterTable` to smaller functions, `alterTable` is about __50__ lines .

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -71,41 +71,31 @@ function mixinMigration(MySQL, mysql) {
 
           if (!err && fields && fields.length) {
             //if we already have a definition, update this table
-            self.alterTable(model, fields, indexes, foreignKeys, function(err, changed, res) {
-              //check to see if there were any new foreign keys for this table.
-              //If so, we'll create them once all tables have been updated
-              if (!err && res && res.newFks && res.newFks.length) {
-                foreignKeyStatements.push(res.newFks);
+            self.alterTable(model, fields, indexes, foreignKeys, function(err, result) {
+              if (!err) {
+                self.addForeignKeys(model, function(err, result) {
+                  done(err);
+                });
+              } else {
+                done(err);
               }
-              done(err);
             });
           } else {
             //if there is not yet a definition, create this table
-            var res = self.createTable(model, function(err) {
+            self.createTable(model, function(err) {
               if (!err) {
-                //get a list of the alter statements needed to add the defined foreign keys
-                var newFks = self.getForeignKeySQL(model, foreignKeys);
-
-                //check to see if there were any new foreign keys for this table.
-                //If so, we'll create them once all tables have been updated
-                if (newFks && newFks.length) {
-                  foreignKeyStatements.push(self.getAlterStatement(model, newFks));
-                }
+                self.addForeignKeys(model, function(err, result) {
+                  done(err);
+                });
+              } else {
+                done(err);
               }
-              done(err);
             });
           }
         });
       });
     }, function(err) {
-      if (err) return cb(err);
-
-      //add any new foreign keys
-      async.each(foreignKeyStatements, function(addFkStmt, execDone) {
-        self.execute(addFkStmt, execDone);
-      }, function(err) {
-        cb(err);
-      });
+      return cb(err);
     });
   };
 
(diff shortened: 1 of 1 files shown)


ssh24:
@slnode test please

ssh24:
@slnode test please

lib/migration.js
```diff
@@ -291,6 +248,34 @@ function mixinMigration(MySQL, mysql) {
...
+    return sql;
+  };
+
+  MySQL.prototype.addIndexes = function(model, actualIndexes, cb) {
+    var self = this;
+    var m = self._models[model];
+    var propNames = Object.keys(m.properties).filter(function(name) {
+      return !!m.properties[name];
+    });
+    var indexNames = m.settings.indexes && Object.keys(m.settings.indexes).filter(function(name) {
```
jannyHou: there is a dup code in `alterTable`, L518, better to remove it :)

lib/migration.js
```diff
@@ -412,92 +397,162 @@ function mixinMigration(MySQL, mysql) {
...
+        if (err) cb(err);
+        else
+          cb(null, result);
+      });
+    } else cb(null, {});
+  };
+
+  MySQL.prototype.dropForeignKeys = function(model, actualFks, cb) {
+    var self = this;
+    var m = self._models[model];
```
jannyHou: I think use the original `var m = this.getModelDefinition(model)` would be better, `getModelDefinition` is defined here:
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L87
From the implementation they may return a different thing

lib/migration.js
```diff
@@ -248,40 +188,57 @@ function mixinMigration(MySQL, mysql) {
...
+    var self = this;
+    sql = sql.concat(self.getColumnsToDrop(model, actualFields));
+    return sql;
+  };
 
+  MySQL.prototype.getColumnsToDrop = function(model, actualFields) {
+    var self = this;
+    var fields = actualFields;
+    var sql = [];
+    var m = self._models[model];
```
jannyHou: same here

lib/migration.js
```diff
@@ -166,80 +156,30 @@ function mixinMigration(MySQL, mysql) {
...
-      return !!m.properties[name];
-    });
-    var indexes = m.settings.indexes || {};
-    var indexNames = Object.keys(indexes).filter(function(name) {
-      return !!m.settings.indexes[name];
-    });
-
-    //add new foreign keys
-    var correctFks = m.settings.foreignKeys || {};
+    var m = self._models[model];
```
jannyHou: same here.

lib/migration.js
```diff
@@ -412,92 +397,162 @@ function mixinMigration(MySQL, mysql) {
...
 
-    function expectedColNameForModel(propName, modelToCheck) {
-      var mysql = modelToCheck.properties[propName].mysql;
-      if (typeof mysql === 'undefined') {
-        return propName;
-      }
-      var colName = mysql.columnName;
-      if (typeof colName === 'undefined') {
-        return propName;
+    var propNames = Object.keys(m.properties).filter(function(name) {
```
jannyHou: same here. dup code. each sub function will define `propNames` by itself.

lib/migration.js
```diff
@@ -412,92 +397,162 @@ function mixinMigration(MySQL, mysql) {
         }
       }
     });
+    if (sql.length > 0) {
+      self.applySqlChanges(model, sql, function(err, result) {
```
jannyHou: Compared of old method `getAlterStatement`, the refactored `applySqlChanges` doesn't do a "join" of sql, while actually here `sql` is `sqls`, which is an array of `sql`.

I will double check is `applySqlChanges` aimed to apply a single sql statement or join and apply multiple ones.

Old method `getAlterStatement`:
```js
MySQL.prototype.getAlterStatement = function(model, statements) {
    return statements.length ?
      'ALTER TABLE ' + this.tableEscaped(model) + ' ' + statements.join(',\n') :
      '';
  };
```

New `applySqlChanges`
```js
MySQL.prototype.applySqlChanges = function(model, pendingChanges, cb) {
    var self = this;
    var thisQuery = 'ALTER TABLE ' + self.tableEscaped(model) + ' ' + pendingChanges.toString();
    thisQuery = thisQuery + ';';
    if (pendingChanges.length) {
      self.execute(thisQuery, cb);
    }
  };
```

lib/migration.js
```diff
@@ -412,92 +397,162 @@ function mixinMigration(MySQL, mysql) {
         }
       }
     });
+    if (sql.length > 0) {
```
jannyHou: The old function has a `checkOnly`, if missing `done` then just check but don't execute the sql.
Now the var is defined in L508 but not checked in this function.

lib/migration.js
```diff
@@ -412,92 +397,162 @@ function mixinMigration(MySQL, mysql) {
...
-  };
+    };
 
-  MySQL.prototype.getAlterStatement = function(model, statements) {
-    return statements.length ?
-      'ALTER TABLE ' + this.tableEscaped(model) + ' ' + statements.join(',\n') :
-      '';
+    async.series([
+      function(cb) {
+        applyPending([self.getAddModifyColumns], cb);
```
jannyHou: In postgresql, items returned from `getAddModifyColumns` and `getDropColumns`   are executed one by one, while here they are executed parallel, not sure if that will cause problem.

ssh24: @jannyHou I disagree. `async.series` invokes functions serially (waiting for each preceding one to finish before starting next), hence it will execute `getAddModifyColumns` first and then `getDropColumns`.

lib/migration.js
ssh24: I don't quite understand what you mean

lib/migration.js
jannyHou: @ssh24 ah my bad, I thought it's async.each. Then it's good :)

Issue #250, opened 2017-03-09
Properties with mysql custom "columnName" don't get autoupdated


stelescuraul:
### Bug or feature request
If you have a model that has properties with `mysql` field and `columnName` that is different than the model property name, then it will throw error on autoupdate

- [x] Bug
- [ ] Feature request

### Description of feature (or steps to reproduce if bug)
Say we have a model "Users" with the following property: 
```
"firstName": {
      "type": "String",
      "required": true,
      "length": 255,
      "precision": null,
      "scale": null,
      "mysql": {
        "columnName": "first_name",
        "dataType": "varchar",
        "dataLength": 255,
        "dataPrecision": null,
        "dataScale": null,
        "nullable": "N"
      },
      "_selectable": true
    }
```
Because it is save in the database as "first_name" and not "firstName" if you change the "nullable" property to "T" (true) the mysql connector throws error when doing autoupdate: 

`Error: ER_BAD_FIELD_ERROR: Unknown column 'firstName' in 'modelName'`

### Expected result
To be able to detect that the property name has a different name in the database, and use that when  you call autoupdate function on the datasource.

### Fix 
I have followed down the flow and I managed to fix this error changing the line 310 in lib/migration.js to use the mysql property name if it is found as follows: 
```
   var pName;
   if(newSettings.mysql && newSettings.mysql.columnName) pName = newSettings.mysql.columnName;
   else pName = self.client.escapeId(propName);
```

the above code replaces line 310. Using that, the mysql connector works fine when doing autoupdate on an existing database. 

I have run the tests and they pass as expected. 

### Additional information (Node.js version, LoopBack version, etc)
```connector version: 2.4.1 
node version : 7.7.1
strongloop v6.0.3 (node v7.7.1)
loopback version: 2.38.0```

ygaller:
Great catch! Ran into the same problem myself with loopback-connector-mysql@2.4.1

For a more succinct version one could just change it to this:
`var pName = (newSettings.mysql && newSettings.mysql.columnName) || self.client.escapeId(propName);`

ssh24:
@stelescuraul Could you please submit a PR to fix this patch? Please tag me on it and I will take a look at it.

ssh24:
@stelescuraul @ygaller Any updates on submitting a PR for this patch?

stelescuraul:
@ssh24 will do it in an hour.

stelescuraul:
@ssh24 Please check PR.

[...]

PR #255, opened 2017-04-03
Fixed the different column name in datasource

stelescuraul:
Issue #250

connect to https://github.com/strongloop/loopback-connector-mysql/issues/250

### Description
 Fixed the issue with autoupdating caused by different columns specs.

#### Related issues

- #250 

### Checklist

I found no need to create new tests for this. If necessary, I will but please give me a few info.

### DIFF
--- a/test/mysql.autoupdate.test.js
+++ b/test/mysql.autoupdate.test.js
@@ -222,12 +222,6 @@ describe('MySQL connector', function() {
 
   it('should update the nullable property of "first_name" to false', function(done) {
     // update the model "required" property
-    updateSchemaColRenameTest();
-    // nullable should be updated to false
-    updateProperty(done);
-  });
-
-  function updateSchemaColRenameTest() {
     var schema = {
       name: 'ColRenameTest',
       options: {
@@ -257,9 +251,8 @@ describe('MySQL connector', function() {
     };
 
     ds.createModel(schema.name, schema.properties, schema.options);
-  }
 
-  function updateProperty(done) {
+    // nullable should be updated to false
     ds.autoupdate('ColRenameTest', function(err) {
       assert.ifError(err);
       ds.discoverModelProperties('col_rename_test', function(err, props) {
@@ -268,7 +261,7 @@ describe('MySQL connector', function() {
         done();
       });
     });
-  }
+  });
 
   function verifyMysqlColumnNameAutoupdate(done) {
     ds.autoupdate('ColRenameTest', function(err) {
--- a/test/mysql.autoupdate.test.js
+++ b/test/mysql.autoupdate.test.js
@@ -18,99 +18,97 @@ describe('MySQL connector', function() {
   });
 
   it('should auto migrate/update tables', function(done) {
-    var schema_v1 =
-      {
-        'name': 'CustomerTest',
-        'options': {
-          'idInjection': false,
-          'mysql': {
-            'schema': 'myapp_test',
-            'table': 'customer_test',
-          },
-          'indexes': {
-            'name_index': {
-              'keys': {
-                'name': 1,
-              },
-              'options': {
-                'unique': true,
-              },
+    var schema_v1 = {
+      'name': 'CustomerTest',
+      'options': {
+        'idInjection': false,
+        'mysql': {
+          'schema': 'myapp_test',
+          'table': 'customer_test',
+        },
+        'indexes': {
+          'name_index': {
+            'keys': {
+              'name': 1,
+            },
+            'options': {
+              'unique': true,
             },
           },
         },
-        'properties': {
-          'id': {
-            'type': 'String',
-            'length': 20,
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

ssh24:
@stelescuraul Thank you for the contribution. Could you please add some unit tests to go with your patch?

stelescuraul:
@ssh24 I can do that tomorrow. Could you please give me a small set of instructions about the flow of tests that you have? Should I set up a new schema and a model and work from there ?

ssh24:
Thank you. Please take a look at [mysql.autoupdate.test.js](https://github.com/strongloop/loopback-connector-mysql/blob/master/test/mysql.autoupdate.test.js). Add the test there and follow the same convention.

ssh24:
@stelescuraul Any updates on adding the test suite?

ssh24:
Hi @stelescuraul, I would like to make progress with this PR. Any updates on adding the test cases as mentioned above?

loay:
@slnode test please

loay:
please run this command in your terminal as well: 
```./node_modules/.bin/eslint --fix .``` to fix the linting errors. Thanks.

**Comment by stelescuraul** [comment]
@ssh24 I have totally forgot about this, sorry. I will work on the tests this weekend. 
@loay Will do

**Comment by stelescuraul** [comment]
@ssh24 I have added the test. It is only one but it does test what it should. It updates the `required` property and it will update the `nullable` property in ddl as well. 

@loay I have run the command for linting, but somehow it still fails. I checked the log and it says something about line 3 having too many characters. Can you point me in the right direction here ?

File: test/mysql.autoupdate.test.js [hunk 1]
```diff
@@ -222,6 +220,56 @@ describe('MySQL connector', function() {
     verifyMysqlColumnNameAutoupdate(done);
   });
 
+  it('should update the nullable property of "first_name" to false', function(done) {
+    // update the model "required" property
+    updateSchemaColRenameTest();
```
ssh24: Instead of creating two functions and calling them inside the test, could you move the logic on the function in the test case itself?

ssh24:
@stelescuraul Thank you for the contribution! I have left a comment on changing the test case to include the logic of the function itself rather than calling those two functions (it would be a good decision to do the way you did if the functions were called in more than one test case). In addition I have also fixed your commit linter issue.

Could you also submit a similar patch for the latest 3.x branch?

ssh24:
@slnode ok to test

stelescuraul:
@ssh24 I have moved the code form function `updateSchemaColRenameTest` into the test itself. I don't know if you wanted me to move `updateProperty` too, but I have kept it as a function for readability. (I looks more readable in my opinion, but if you`d like, I can move that too).

Also, I can see that i have the same linter problems even after running the command above. Mind sharing with me what the actual problem is ? It seems that the commit message is too long ?

```
**
**  Linting commit logs
**
**  3 problems found:
**    e681494 - Merge remote-tracking branch 'origin/Fix/automigrat: First line should be 50 characters or less (saw 88)
**    6a81d31 - moved the code from the function into the test bloc: First line should be 50 characters or less (saw 52)
**    1627aa9 - Added test to change the nullable property.: line 3 longer than 72 characters.
**
```

I will do a fix for 3.x tomorrow.

test/mysql.autoupdate.test.js
```diff
@@ -222,6 +220,53 @@ describe('MySQL connector', function() {
...
+      },
+    };
+
+    ds.createModel(schema.name, schema.properties, schema.options);
+
+    // nullable should be updated to false
+    updateProperty(done);
+  });
+
+  function updateProperty(done) {
```
ssh24: @stelescuraul Yes, please move this function into the test case itself. Again I like your idea to put long logics itself into a function but not worthy of it if only one test case uses it.

stelescuraul: @ssh24 done. Can you give me a hint on those linting issues ?

ssh24:
@stelescuraul Thank you for the fixes. Yes, so there are some commit linter rules that we follow. Here is the complete breakdown of the rules:
```
1. The first line of the commit message cannot be no longer than 50 characters.
2. Second line should be always empty.
3. Every consecutive line after the second line should be no longer than 72 characters long.
```
I have fixed your commit linter error again. So please pull from your branch to get the latest.

stelescuraul:
@ssh24 Alright. noted :smiley: I will have the fix for the 3.x tomorrow. Should I create a new issue or just reference the existing one in a new pr ?

ssh24:
@stelescuraul Sounds good. You can refer it to the same issue #250. Please let me know when that PR is ready and I will look into it. :)

ssh24:
@slnode test please

Issue #149, opened 2016-02-11
Date fields - timezone problem


euronymous:
Hi,

I've spend the last hour for reverse engineering of that case...

Lets look at the file:
/loopback-connector-mysql/lib/mysql.js:353
        val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
Why you don't give a sh about timezone?! This is at least weird.

Here is an example. I have a date from MySQL:
Mon Jun 01 2015 02:01:01 GMT+0200 (CEST)

So lets try:

```
        console.log(val);
        console.log(val.toString().replace(/GMT.*$/, 'GMT'));
        val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
        console.log(val);
```

And the result is:

```
        Mon Jun 01 2015 02:01:01 GMT+0200 (CEST)
        Mon Jun 01 2015 02:01:01 GMT
        Mon Jun 01 2015 04:01:01 GMT+0200 (CEST)
```

And at last the strongloop API response:
"2015-06-01T02:01:01.000Z"

Kind Regards,
Krzysiek

bbito:
@euronymous I had an issue with non UTC dates in MySQL and found the dateString option useful for a workaround. It may help in your case as well - see issue 120 and daddybh's dateString option: https://github.com/strongloop/loopback-connector-mysql/issues/120

DaGaMs:
I also just ran into this problem when writing some tests where I entered data into a mysql database directly, then fetched it with loopback. My understanding is that `DATETIME` fields are "timezone-free". Most libraries, including e.g. [mysqljs](https://github.com/mysqljs/mysql#connection-options) interpret `DATETIME` objects in the local timezone by default. Loopback however turns them into UTC by default, which is a bit confusing.

bbito:
@DaGaMs , Since this is a JavaScript/node system, I think this is the correct behavior for DATETIME and TIMESTAMP columns since JavaScript serializes its DATE object into ISO 8601 'which is always 24 characters long: YYYY-MM-DDTHH:mm:ss.sssZ. The timezone is always zero UTC offset, as denoted by the suffix "Z".' See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON

superkhau:
[...]

PR #257, opened 2017-04-09
refactor date, timestamp and datetime data types handling

darknos:
This is a continue of #170.

I  propose more correct way (from my point of view) of handling DATE, TIMESTAMP and DATETIME types.

1. Prevent double converting of date strings received from mysql server to js Date object using dateStrings: true setting for mysql connection.
2. Add support for timezone settings for cases when MysqlServer timezone is different from local one.
3. Add different handling of DATETIME and TIMESTAMP types as described in mysql documentation 
> MySQL converts TIMESTAMP values from the current time zone to UTC for storage, and back from UTC to the current time zone for retrieval. (This does not occur for other types such as DATETIME.) By default, the current time zone for each connection is the server's time. The time zone can be set on a per-connection basis. As long as the time zone setting remains constant, you get back the same value you store. If you store a TIMESTAMP value, and then change the time zone and retrieve the value, the retrieved value is different from the value you stored. This occurs because the same time zone was not used for conversion in both directions. The current time zone is available as the value of the time_zone system variable.
4. Correct way of handling DATE type 
5. Ability to work with DATE, TIMESTAMP and DATETIME fields without any conversion (type:String instead of type:Date in property definition)
6. Added unit tests to check all possible cases, different time zones, and date types.

#### Related issues
#170
### Checklist

connected to https://github.com/strongloop-internal/scrum-apex/issues/173


darknos:
ping @ssh24

slnode:

slnode:

slnode:

slnode:

ssh24:
@slnode test please

test/datetypes.notz.test.js
```diff
@@ -0,0 +1,186 @@
...
+
+
+function getTimeZone() {
+  var offset = new Date().getTimezoneOffset(), o = Math.abs(offset);
+  return (offset < 0 ? "+" : "-") + ("00" + Math.floor(o / 60)).slice(-2) + ":" + ("00" + (o % 60)).slice(-2);
+}
+
+var localTZ = getTimeZone();
+
+describe('MySQL DATE, DATETTIME, TIMESTAMP types on server with local TZ', function() {
```
ssh24: Please move these two test cases all under one test file. For example in [./test/datatypes.test.js/](https://github.com/strongloop/loopback-connector-mysql/blob/master/test/datatypes.test.js).

test/datetypes.notz.test.js
```diff
@@ -0,0 +1,186 @@
...
+  before(setup);
+
+  it('should set local timezone in mysql '+localTZ, function(done) {
+      query("SET @@session.time_zone = '"+localTZ+"';", function(err) {
+          assert.ok(!err);
+          done();
+      });
+  });
+
+  it('should create table', function(done) {
```
ssh24: Don't have a test case for it. And instead of creating a table like this, create a loopback model directly and automigrate the model to the database. More on automigrate [here](https://loopback.io/doc/en/lb2/Creating-a-database-schema-from-models.html#auto-migrate).

test/datetypes.notz.test.js
```diff
@@ -0,0 +1,186 @@
...
+      });
+  });
+
+  it('should disconnect when done', function(done) {
+    db.disconnect();
+    done();
+  });
+});
+
+function setup(done) {
```
ssh24: Don't need any of these setup AFAIK. All the models/tables should be on the default database the test runs against.

test/datetypes.notz.test.js
darknos: I create table via SQL to have a test field with DEFAULT CURRENT_TIMESTAMP:
`timestampDefaultField timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,`.

for this test case:

```
  it('timestampField shoud equal DEFAULT CURRENT_TIMESTAMP field', function(done) {
      DateModel.findOne({where: {id: 1}}, function(err, found) {
        assert.ok(!err);
        assert.equal(found.timestampField.toISOString(), found.timestampDefaultField.toISOString());
        done();
      });
  });
```
It is needed to be sure that time set on server side is the same as time we pass from new Date();

I know how to use auto migration but don't know how to create field with such defaults. I can do auto migrate and then query `ALTER TABLE CHANGE timestampDefaultField timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP`. Is it acceptable?

darknos: and, sure, I'll move it to setup section instead of test case

test/datetypes.notz.test.js
darknos: I unified setup procedures for both test cases

darknos:
Done. @ssh24, please review.

ssh24:
@darknos Looks like loopback does not support `timestamp` and `datetime` as a data type as of yet. Look at the supported data types [here](https://loopback.io/doc/en/lb2/LoopBack-types.html). Those are specific to mysql and I am not sure if it is in our roadmap.

@raymondfeng Could you correct me if I am wrong?

darknos:
Js Date type is mapped to mysql datetime, date or timestamp. 

I use loopack on tens of clients project and in most cases in their databases were these three types. loopback has to support all of them.

test/date_types.test.js
```diff
@@ -0,0 +1,251 @@
...
+    done();
+  }); });
+
+var prepareModel = function(tz, migrate, done) {
+  db = getSchema({timezone: tz});
+  DateModel = db.define('DateModel', {
+    id: {type: Number, id: 1, generated: true},
+    datetimeField: {type: Date, dataType: 'datetime', null: false},
+    timestampField: {type: Date, dataType: 'timestamp', null: false},
+    timestampDefaultField: {type: Date, dataType: 'timestamp', null: false},
```
ssh24: @darknos You would need to change the `dataType` in that case here then. LoopBack does not support `timestamp` as an official data type. We support `date` which is the JavaScript date object. More info on [docs](https://loopback.io/doc/en/lb2/LoopBack-types.html).

Same for `datetime` as a data type.

darknos: https://github.com/strongloop/loopback-connector-mysql/blob/master/README.md#loopback-to-mysql-types

>Type mappings:
>LoopBack to MySQL types
>Date -> DATETIME

>MySQL to LoopBack types
>DATE,TIMESTAMP,DATETIME  -> Date

I just want to make this mapping correct. Try my simple and clear test cases with current version of connector to see what is wrong.

test/date_types.test.js
```diff
@@ -0,0 +1,261 @@
...
+  });
+
+  it('should find model instance by datetime field', function(done) {
+    DateModel.findOne({
+      where: {
+        datetimeField: date,
+      },
+    }, function(err, found) {
+      assert.ok(!err);
+      assert.ok(found);
```
ssh24: Could assert for something more than just found?

test/date_types.test.js
```diff
@@ -0,0 +1,261 @@
...
+  });
+
+  it('should find model instance by timestamp field', function(done) {
+    DateModel.findOne({
+      where: {
+        timestampField: date,
+      },
+    }, function(err, found) {
+      assert.ok(!err);
+      assert.ok(found);
```
ssh24: Same here. Assert more than just `found`

test/date_types.test.js
```diff
@@ -0,0 +1,261 @@
...
+  });
+
+  it('should find model instance by date field', function(done) {
+    DateModel.findOne({
+      where: {
+        dateField: dateOnly,
+      },
+    }, function(err, found) {
+      assert.ok(!err);
+      assert.ok(found);
```
ssh24: Same here. Assert more than just `found`

lib/mysql.js
```diff
@@ -356,7 +385,15 @@ MySQL.prototype.toColumnValue = function(prop, val) {
...
       val = new Date(val);
     }
-    return dateToMysql(val);
+
+    if (prop.dataType == 'date') {
+      return dateToMysql(val).substring(0, 10);
+    } else if (prop.dataType == 'timestamp' || (prop.mysql && prop.mysql.dataType == 'timestamp')) {
+      var tz = this.client.config.connectionConfig.timezone;
+      return dateToMysql(val, tz);
+    } else {
```
ssh24: Remove the else. Just `return dateToMysql(val);` is good.

lib/mysql.js
```diff
@@ -307,17 +310,43 @@ MySQL.prototype.updateOrCreate = function(model, data, options, cb) {
...
     return v < 10 ? '0' + v : v;
   }
+
+  function convertTimezone(tz) {
+    if (tz === 'Z') {
+      return 0;
+    }
+
+    var m = tz.match(/([\+\-\s])(\d\d):?(\d\d)?/);
+    if (m) {
```
ssh24: One liner for this should be fine: `if (m) return (m[1] == '-' ? -1 : 1) * (parseInt(m[2], 10) + ((m[3] ? parseInt(m[3], 10) : 0) / 60)) * 60;`

lib/mysql.js
```diff
@@ -307,17 +310,43 @@ MySQL.prototype.updateOrCreate = function(model, data, options, cb) {
...
+    return dt.getFullYear() + '-' +
+      fillZeros(dt.getMonth() + 1) + '-' +
+      fillZeros(dt.getDate()) + ' ' +
+      fillZeros(dt.getHours()) + ':' +
+      fillZeros(dt.getMinutes()) + ':' +
+      fillZeros(dt.getSeconds());
+  } else {
+    tz = convertTimezone(tz);
+
+    if (tz !== false && tz !== 0) {
```
ssh24: Change this to: `if (tz !== false && tz !== 0) dt.setTime(dt.getTime() + (tz * 60000));`

test/date_types.test.js
darknos: `assert.equal(found.id, 1);` should be enough - we have to be sure that param of each type is working and found proper recored.

darknos:
Done. @ssh24 please review

test/date_types.test.js
```diff
@@ -95,7 +97,19 @@ describe('MySQL DATE, DATETTIME, TIMESTAMP types on server with local TZ', funct
       },
     }, function(err, found) {
       assert.ok(!err);
-      assert.ok(found);
+      assert.equal(found.id, 1);
+      done();
+    });
+  });
+
+  it('should find model instance by date field passing string', function(done) {
```
ssh24: @darknos Test is failing locally for this. Why is this a valid test case to test for a `dateField` of type string when it expects a Date object?

test/date_types.test.js
```diff
@@ -202,6 +217,18 @@ describe('MySQL DATE, DATETTIME, TIMESTAMP types on server with non local TZ (+0
     });
   });
 
+  it('should find model instance by date field passing string', function(done) {
```
ssh24: Same reason here. Test is failing locally for this. Why is this a valid test case to test for a `dateField` of type string when it expects a Date object?

test/date_types.test.js
darknos: All "where" parameters are passed normalize function of "juggler". 
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1146
If property is one of type Date it does new Date(arg)
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1540

for DATE type test case will pass for timezones like my one (+xxxx):
"2016-12-22" -> "Thu Dec 22 2016 02:00:00 GMT+0200 (EET)"
but 
will fail for (-xxxx) timezones:
"2016-12-22" -> "Thu Dec 21 2016 22:00:00 GMT-0200"

I'm going to remove this test case but I'd suggest to mean it in documentation.

darknos: It is a question how to pass where parameter to DATE field in different timezone via REST interface. 

f.e. server is in GMT-6000 timezone and client is in GMT+2000 timezone. on client side we do:  var filter = {where:{dateFiled:new Date(2016,11,22)}};  it will pass as {"where":{"dateFiled":"2016-12-21T22:00:00.000Z"}}

and dao will convert it to date "Wed Dec 21 2016 16:00:00 GMT-0600 (CST)" and pass to SQL as "2016-12-21"

darknos:
It is a question how to pass where parameter to DATE field in different timezones via REST interface.

f.e. server is in `GMT-6000` timezone and client is in `GMT+2000` timezone. on client side we do:
```
  var filter = {where:{dateFiled:new Date(2016,11,22)}};  
```
it will pass via HTTP as `{"where":{"dateFiled":"2016-12-21T22:00:00.000Z"}}`
and dao will convert it to date `Wed Dec 21 2016 16:00:00 GMT-0600 (CST)` and pass to SQL as `2016-12-21` and we get wrong result. :(

ssh24:
@darknos Not sure what you mean
> it will pass via HTTP as {"where":{"dateFiled":"2016-12-21T22:00:00.000Z"}}
and dao will convert it to date Wed Dec 21 2016 16:00:00 GMT-0600 (CST) and pass to SQL as 2016-12-21 and we get wrong result

What is the wrong result?

bbito:
I wish I had time right now to really look at the implications of this PR, but unfortunately I do not right now. I do want to comment that I'm concerned about forcing the mysql config: `dateStrings: true` since users could previously assume that mysql config settings in their datasources.json would be passed to the underlying mysql module. 
[https://github.com/strongloop/loopback-connector-mysql/pull/257/files#diff-fd1940159a592c074df64da530d394e3R140](url)
I also want to point to an issue I was involved with that I feel is closely related: [https://github.com/strongloop/loopback-connector-mysql/issues/149](url)
That issue identifies that this connector currently does not properly handle MySQL timezone settings and in fact currently stores and retrieves UTC. A fix is a breaking change because datetime comparisons between data created before and after such a fix will be incorrect because of the long-standing behavior of storing UTC despite the docs stating that the default is 'local'.

darknos:
>@darknos Not sure what you mean

>>it will pass via HTTP as {"where":{"dateFiled":"2016-12-21T22:00:00.000Z"}}
and dao will convert it to date Wed Dec 21 2016 16:00:00 GMT-0600 (CST) and pass to SQL as 2016-12-21 and we get wrong result

>What is the wrong result?

Pretend we have field dateField with value "2016-12-22" stored in database and I want to get records with this value. I'll do
```
var filter = {where:{dateFiled:new Date(2016,11,22)}};  
```

but it will execute SQL `select * from table where dateField = ' 2016-12-21'`

darknos:
@bbito 
>That issue identifies that this connector currently does not properly handle MySQL timezone settings and in fact currently stores and retrieves UTC. A fix is a breaking change because datetime comparisons between data created before and after such a fix will be incorrect because of the long-standing behavior of storing UTC despite the docs stating that the default is 'local'.

this PR fixes timezone handling and makes difference between handling DATETIME and TIMESTAMP as described in MySQL documentation. 

I found only one way to fix it - make only one conversion from mysql strings to js Date object and do it in the loopback connector but not on node-mysql side.

Probably there are another ways to fix it but I'd ask to keep test/date_types.test.js I made for it and even extend it if you have any ideas about other test cases. I guess you can get ideas of new test cases from https://github.com/strongloop/loopback-connector-mysql/issues/149

bbito:
@darknos 

> makes difference between handling DATETIME and TIMESTAMP as described in MySQL documentation.
I don't think there should be a difference in the way this connector handles DATETIME vs TIMESTAMP
[https://github.com/strongloop/loopback-connector-mysql/issues/149#issuecomment-235946823](url)

ssh24:
Still not sure what is the problem.

> Pretend we have field dateField with value "2016-12-22" stored in database and I want to get records with this value. I'll do
```
var filter = {where:{dateFiled:new Date(2016,11,22)}};  
```
> but it will execute SQL `select * from table where dateField = ' 2016-12-21'`

Why should it execute SQL with dateField's date equal to 21? I am assuming the server is in GMT, and you are accessing it from somewhere GMT-x timezone?

darknos:
@bbito 

>>makes difference between handling DATETIME and TIMESTAMP as described in MySQL documentation.

>I don't think there should be a difference in the way this connector handles DATETIME vs TIMESTAMP:
>#149 (comment)

There is a simple way to see that the difference is present. It is described in my test cases.
1. add the same value to DATETIME and TIMESTAMP field
2. change @session.time_zone value
3. retrieve values and mysql will return different values for these fields

```
mysql> create table test (dt  datetime, ts timestamp);
Query OK, 0 rows affected (0.01 sec)

mysql> insert into test values ('2017-04-11 14:38:07', '2017-04-11 14:38:07');
Query OK, 1 row affected (0.00 sec)

mysql> select * from test;
+---------------------+---------------------+
| dt                  | ts                  |
+---------------------+---------------------+
| 2017-04-11 14:38:07 | 2017-04-11 14:38:07 |
+---------------------+---------------------+
1 row in set (0.00 sec)

mysql> set @@session.time_zone = '+05:30';
Query OK, 0 rows affected (0.00 sec)

mysql> select * from test;
+---------------------+---------------------+
| dt                  | ts                  |
+---------------------+---------------------+
| 2017-04-11 14:38:07 | 2017-04-11 17:08:07 |
+---------------------+---------------------+
1 row in set (0.00 sec)
```

Just runt test/date_types.test.js with DEBUG=*mysql* to see mysql data packets.

darknos:
>Still not sure what is the problem.

>>Pretend we have field dateField with value "2016-12-22" stored in database and I want to get records with this value. I'll do
>>var filter = {where:{dateFiled:new Date(2016,11,22)}};  
>>but it will execute SQL select * from table where dateField = ' 2016-12-21'

>Why should it execute SQL with dateField's date equal to 21? I am assuming the server is in GMT, and you are accessing it from somewhere GMT-x timezone?

1. in GMT+2 tz `new Date(2016,11,22)` will be converted to `"2016-12-21T22:00:00.000Z"`
2. when loopback receive rest route to `find` it will "normalize" where statement 
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1146
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1540
3. if server is in GMT+0 - no problem. we'll get `Wed Dec 22 2016 00:00:00 GMT` and in loopback-connector-mysql, `toColumnValue` function will convert it to string `2016-12-22` and pass to mysql.
4. BUT if server is not in UTC. (f.e. GMT-05:00 like when you started "strings" test case on your side) it will convert "2016-12-21T22:00:00.000Z" to `Wed Dec 21 2016 00:00:00 GMT` and `toColumnValue` function will convert it to string `2016-12-21`.

this is not issue but I'd suggest to describe it as "notice" in documentation.

ssh24:
> 4. BUT if server is not in UTC. (f.e. GMT-05:00 like when you started "strings" test case on your side) it will convert "2016-12-21T22:00:00.000Z" to `Wed Dec 21 2016 00:00:00 GMT` and `toColumnValue` function will convert it to string `2016-12-21`.

Could we add a check to see the timezone and if the timezone is so ahead/behind that it falls into a new day we automatically falls back to use the new day?

For example:
Let's say we have two server timezones. Tz1 = GMT +7 and Tz2= GMT -7

1. With Tz1, let's say we want to convert `"2016-12-22T05:00:00.000Z"`. It should return `Wed Dec 21 2016 22:00:00 GMT` and the `toColumnValue` function upon conversion should check the date we want to convert versus the one mysql returns (22 vs 21) and if they are different by +/-1 we should use the +/-1 value and convert it to string f.e on this case it should be `2016-12-22` and not `2016-12-21`.

2. Same case for Tz2. If we want to convert `"2016-12-22T22:00:00.000Z"`, it should return `Fri Dec 23 2016 05:00:00 GMT` and the `toColumnValue` function upon conversion should check the date we want to convert versus the one mysql returns (22 vs 23). Since the date are different, we use the actual date and convert it to string f.e on this case it should be `2016-12-22` and not `2016-12-23`.

darknos:
Conversion of where parameter is not in mysql connector `toColumnValue` but in juggler/lib/dao.js module. 

`Model.find({where:{dateField:value}})` goes completely out of mysql connector but in base connector. 
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1540

I'd suggest just add notice with "Known possible issue" to README.md

darknos:
I'd propose simple way how to work with DATE field.

1. if you want to work with DATE field in mysql - always pass UTC date to where or create/update
```javascript
new Date(Date.UTC(2016,11,22));
//instead of
new Date(2016,11,22);
```

and in `toColumnValue`
```javascript
val = val.toISOString().substring(0,10);
```

darknos:
but in case above we need return Date field in UTC too (in `fromColumnValue`)
and this can be confusing. :( and it differs from node-mysql type casting.

so idea above is bad. :(

darknos:
I feel that the best way to work with DATE field is:
```javascript
dateField: {type:"String", dataType:"DATE"}
```
and it not a joke :(

ssh24:
@darknos Could you update the docs with the possible issue you said?

@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

darknos:
I'm not so good technical writer. Check is theft below clear enough:

### Date value types

For TIMESTAMP and DATE types, use the `dateType` option to specify custom type. Default is DATETIME.

Example:

```javascript
{ created :
  { type: Date, 
    dataType: 'timestamp'
  }
}
```

Notice: Quering DATE type, remember that values sent to server via REST will be converted to Date object using server timezone and only YYYY-MM-DD part will be used in SQL query. For example if a client in GMT+2 timezone and server in GMT-2 timezone:

```javascript
$scope.products = Product.find({where:{expired:new Date(2016,11,22)}});
```

will be converted to URL like: `/api/Products/?filter={"where":{"expired":"2016-12-21T22:00:00Z"}}`

and on server with timezone GMT-2 it becames `2016-12-21 20:00:00 GMT-0200` and SQL will be like this:

```SQL
SELECT * FROM Product WHERE expired = '2016-12-21'
```
which is not expected behavior.

Another option to work with DATE field and avoid timezone problem is use String type. 

```javascript
{ birthday :
  { type: String, 
    dataType: 'date'
  }
}
```

ssh24:
@slnode test please

ssh24:
@darknos Updated the readme with your comments. Thank you for the contribution! PR should be ready to land once CI passes.

@slnode test please

DaGaMs:
Just a very quick note to say that this change breaks all sorts of things for me - I've not had a chance to even really understand it fully, but the fact that this PR seems to completely change the behaviour of the API is a pretty serious issue!

bbito:
As a github newb, I'm not sure of the protocol here so forgive me if this is bad manners, but I want the folks involved with https://github.com/strongloop/loopback-connector-mysql/issues/149 to have a look at this so I'll call them out here: @siddhipai , @jannyHou, @bajtos, @superkhau
@DaGaMs has already popped over.

DaGaMs:
Yeah, I'm currently trying to write some tests to narrow down what I think is "wrong" - or at least unexpected - with the new behaviour. I think it has to do with how values that are saved in DATETIME columns are handled. I'll post as soon as I've made progress...

bbito:
@DaGaMs even if this works 'correctly', my comment over at https://github.com/strongloop/loopback-connector-mysql/issues/149#issuecomment-294949451 stands because loopback-connector-mysql will now honor the connection or server settings for time zone and all DATETIME and TIMESTAMP fields committed via the previous connector versions will now be returned as offset by the difference between the server/connection setting and UTC.

ssh24:
This is a major jump in our API (not released yet). It fixes the issue with conversion of `DATETIME and TIMESTAMP` to UTC on commit and also not honouring the timezone setting of the MySQL server.

Now thank you to @DaGaMs for bringing up the potential issue with previous versions of the connector returning an offset by the difference between server and UTC setting.

I believe this should be made well aware of for the users using MySql date options in our docs.

DaGaMs:
Well the problem is that this now makes my tests fail in weird and wonderful ways that I don't quite understand yet - hence my comment above that I feel as if something is not quite right, but I can't tell you yet if this is an issue with my tests or with this patch.

In any case, @bbito is right, this will probably screw things up badly for users with existing data in the db

superkhau:
>This is a major jump in our API (not released yet)

@DaGaMsHow are your tests failing differently?

@ssh24 Do you have reference does noting this major change? Can you paste the link here?

superkhau:
@ssh24 Please apply the correct labels next time too.

superkhau:
cc @kjdelisle You can take over from here, I think any breaking changes should not be landed/reverted because it seems @DaGaMs and @bbito are running into issues.

ssh24:
I apologize for completely misunderstanding this PR. Tried out some basic queries with mysql date, datetime and timestamp attached to a loopback model and this PR should not have been landed in first place.

- In a loopback model with a date type
  - Upon insertion of new data it will always commit the date into the corresponding mysql table as a UTC string (no matter if the actual date string is given in UTC timezone or local).
  - Upon retrieval of a date object from the database it will always convert the date back from UTC to local timezone.

LoopBack also does not differentiate between the `date, datetime and timestamp` at the end of the day they are all a `date` type and the mysql driver even converts it into a javascript date time for example `Thu Apr 20 2017 19:29:12 GMT-0400 (EDT)` upon retrieval.

This is the expected behaviour which gives the end-user full control on how dates should be manipulated in the client side.

If @raymondfeng @kjdelisle are okay, I am ready to revert this PR.

I deeply apologize to other contributors who have been affected by this.

darknos:
Hi,

I was an initiator of the changes and it was my fault that I missed a "breaking change" nature of the PR. I used a fixed version  of the connector last two years, so I just forgot about it. 

I have an idea what to do to keep both kinds of behavior:

1. Lets add an option called like `"legacyDateTypeProcessing"` with default `true` and return small piece of code with an old style date types processing
2. If you explicitly set `legacyDateTypeProcessing: false` the new code will work

It gives us a chance to migrate without any headaches.

DaGaMs:
After A LOT of banging my head against the wall, I think I can explain my problem a bit better:

If you create a row in a table with a DATETIME value of `"2017-04-21 12:47:00"`, and you have set the `timezone` of your db connector to *e.g.* `+04:00`, what would you expect to get when you do a `find` of that row in the table?

My thinking is that the timezone setting of the connection should define how the data is interpreted. So I would expect to get back `new Date("2017-04-21 12:47:00+04:00")` in loopback.

Instead, I get `new Date("2017-04-21 12:47:00+04:00")`. This means it depends on the implementation of the JavaScript `Date` what timezone setting you get for DATETIME columns, as far as I can tell. I think that's not a good solution.

I have created a test for this in my branch of the code, see [this commit](https://github.com/DaGaMs/loopback-connector-mysql/commit/ac1a6a05c4078b13bf26d676e378fab45d89e7ec).

DaGaMs:
In fact, this leads to really bizarre behaviour. Most worryingly, it can be totally random if a date is interpreted as UTC or local timezone. I'm in GMT+1, so I get:
```
# interpreted as UTC
> console.log(new Date("2016-01-03 00:08:06"))
2016-01-03T00:08:06.000Z

# interpreted as local
> console.log(new Date("2016-10-23 00:08:06"))
2016-10-22T23:08:06.000Z
```

That's a massive WTF, and cost me some 5 hours ;-)

darknos:
In you test case, due to MySQL documentation, result should be different in dependency of Mysql and javascript timezones. If they are the same you will get the same result for Timestamp and Datetime field. If timezones are different you will get no difference for datetime but offset for timestamp field.

btw, anycase if you want to work with Strings instead of js Date object you can define model as
```javascript
{type:"String", dataType:"datetime"}
```

If you work with loopback as documented - send Date object to the connector and receive Date object from connector, everything works as expected.

If you send "string" to database and receive Date object you should be aware of timezone configuration. 

The main difference between DATETIME and TIMEZONE in Mysql that timezone is a "moment" (auto converted to proper tz) but datetime is a "string" in table without any tz info.

darknos:
Due to mysql documentation:
 
`row in a table with a DATETIME value of "2017-04-21 12:47:00"` depends from client timezone for me (GMT+2) it should be `2017-04-21T12:47:00+0200` and for you (GMT+4)  it should be `2017-04-21T12:47:00+0400`

`row in a table with a TIMESTAMP value of "2017-04-21 12:47:00"` does not depend from client timezone and (if server in UTC) for me (GMT+2) it should be `2017-04-21T11:47:00+0200` and for you (GMT+4) it should be `2017-04-21T08:47:00+0400`

DaGaMs:
I disagree with this.

You interpret the MySQL documentation to mean that DATETIME values *have no timezone*, so it's ok to leave it to `new Date()` in JS to interpret the values. I disagree. I think. DATETIME values should also be interpreted relative to the MySQL connection timezone. The difference between TIMESTAMP and DATETIME is that TIMESTAMP will return different values if you change the connection timezone, whereas DATETIME won't. But I still feel that DATETIME values should have an **explicit** timezone when converting them to a `Date`.

DaGaMs:
This is all the more important, by the way, if you consider the interpretation problems with timezone-less dates that I [described above](#issuecomment-296174027).

darknos:
@DaGaMs

We work with date time as below and I expect such behavior.
```javascript
let date = new Date();
let dateString = date.toISOString();
DateModel.create({
     id: 1.
     datetimeField: date,
}, function(err, res) {
    DateModel.findOne({where:{dateField: date}, function(err, result) {
      //result.id == 1;
   }     
   //or
    DateModel.findOne({where:{dateField: dateString}, function(err, result) {
      //result.id == 1;
   }     
}
```

where do you suggest to specify explicit timezone?

darknos:
I'm not sure that understand 'bizarre behavior'. In my TZ:

```
> console.log(new Date("2016-01-03 00:08:06"))
Sun Jan 03 2016 00:08:06 GMT+0200 (EET)
> console.log(new Date("2016-10-23 00:08:06"))
Sun Oct 23 2016 00:08:06 GMT+0300 (EEST)
```

one hour difference is Daylight Saving Time difference

```
> console.log(new Date("2016-01-03 00:08:06").toISOString())
2016-01-02T22:08:06.000Z
> console.log(new Date("2016-10-23 00:08:06").toISOString())
2016-10-22T21:08:06.000Z
```

DaGaMs:
I suggest that if you initialise a MySQL datasource with `{timezone: "+04:00"}`, then all DATETIME columns in the datasource should be assumed to be in that timezone. So, if I did a raw `SELECT` and got a value of `2016-04-12 11:23:00`, I would assume that this is the same as `2016-04-12 07:23:00 UTC`.

darknos:
@DaGaMs So, do you just suggest to change this

```javascript
//if datatype is timestamp and zimezone is not local - convert to proper timezone
if (tz !== 'local' && (prop.dataType == 'timestamp' || (prop.mysql && prop.mysql.dataType == 'timestamp'))) {
        dateString += ' ' + tz;
}
```
to  this
```javascript
//if   timezone is not local - convert to proper timezone
if (tz !== 'local') {
        dateString += ' ' + tz;
}
```
?

DaGaMs:
yes. And correspondingly also make sure that whatever gets written to the db is converted to the tz of the connection

darknos:
sure:
```
 } else if (prop.dataType == 'timestamp' || (prop.mysql && prop.mysql.dataType == 'timestamp')) {
    var tz = this.client.config.connectionConfig.timezone;
    return dateToMysql(val, tz);
}
return dateToMysql(val);
```
to
```
    return dateToMysql(val, this.client.config.connectionConfig.timezone);
```

darknos:
@DaGaMs  
https://github.com/iPublisherUA/loopback-connector-mysql/commit/fdda36d961db49b643fc06ce769b0b57c87b3f4d

all tests are passed.

DaGaMs:
I had test failures with your branch, strangely. Here is a branch where all tests are passing for me:

https://github.com/DaGaMs/loopback-connector-mysql/commit/f947c1aeb54028ab59e54866298f427adb15251f

bbito:
Should discussion about this (loopback-connector-mysql date / time zone issues) continue here or at https://github.com/strongloop/loopback-connector-mysql/issues/149?

superkhau:
I think @darknos should open up a PR with the legacyBehaviour flag mentioned above and we can reference this issue and #149 to that PR for the fixup and release ASAP so you and @DaGaMs can get their fixes right away with an immediate publish.

cc @ssh24 @kjdelisle

bbito:
Hey @superkhau , I don't have a 'fix' ready to go. My feeling is that this (https://github.com/strongloop/loopback-connector-mysql/pull/257) goes in the opposite direction from what was discussed at https://github.com/strongloop/loopback-connector-mysql/issues/149 where I was suggesting removing all the code converting dates to UTC and just handing off and receiving js Date objects to/from the underlying mysql module. I do have a fork with that stuff removed that works fine in my limited testing, but I don't have any legacy fallback or the docs for upgrading. This (#257) goes completely the other way and forces the underlying module to produce String values for all date types.

superkhau:
@bbito I hear your concern. Maybe the simplest route is to revert #257 and continue the discussion at #149.

My feeling is we need to do:

- Add a compatibility flag allowing users to opt into old behaviour - this will simplify upgrade from the current version of the connector
- Provide documentation on how to upgrade data stored in MySQL from the UTC format to the new format, to allow users to eventually switch off the compat flag.

as suggested by @bajtos at #149 

Then in a subsequent PR after continuing some discussion at #149:

- Change the default behaviour of this connector to match the suggestinos in the MySQL documentation
- Release the changes as a new major version of the connector

Thoughts?

bbito:
@superkhau That sound best to me! I think what we came up with at #149 was a solid plan.

DaGaMs:
@superkhau I'm pretty sure that with the updated code (after @darknos and my discussion yesterday) the "compatibility setting" would simply be to make `"timezone": "+00:00"` the default for all connections. Then all dates going in and out of the db would be interpreted as UTC, AFAICS.

DaGaMs:
@bbito I'm not sure I agree any more with the idea of leaving all date casting to javascript. In that case, a change in the timezone setting of the server that the loopback instance runs on would change the interpretation of the data. I prefer the more "explicit" setting in the datasource.json file, so it's totally clear that all dates that go in and come out of the database are in a specific timezone (and UTC is a *very* sensible default ;-) )

darknos:
@superkhau, @ssh24 , @DaGaMs  I added new PR: #265 and I'll wait for your decision.

ssh24:
Following up on the discussion between @darknos and @DaGaMs, I am okay with the idea of enabling a flag to allow the previous behavior and the new behavior. If the fixes on #265 are sensible, this PR does not need to be reverted.

bbito:
@DaGaMs 

> I prefer the more "explicit" setting in the datasource.json file

I completely agree and this was the crux of https://github.com/strongloop/loopback-connector-mysql/issues/149#issuecomment-235939569. Any fix must honor `"timezone":"..."` in datasources.json unless the proposed legacy compatibility flag is set in which case the module should continue to ignore the timezone set in datasources. I currently have legacy compatibility default to true - so one needs to opt-in to new behavior, but I'm fine with either true or false for default.

ssh24:
Reverting this PR as it is a breaking change for other users. Please follow https://github.com/strongloop/loopback-connector-mysql/issues/149 as we try to come up with a cleaner solution.

darknos:
got it. after all discussions I see that the best way is to keep own fork of the connector. :(

PR #258, opened 2017-04-10
Increase test timeout

ssh24:


rmg:
The tests are failing due to too many open connections to mysql. Since the tests are run against a per-build server with no other clients, this may be an indication of a connection leak in the tests.

ssh24:
Closing this. Not needed.

PR #262, opened 2017-04-18
Fix async.each in migration

DaGaMs:
### Description

The problem is that FK constraints require a certain table to exist, otherwise the SQL will fail.
I switched from `each` to `eachSeries` so that at least, I can influence the order in wich tables
are processed manually in my migration boot script.

A better solution would be to first finish creating tables, then deal with foreign keys separately.
This would also fix the issue of circular references (TableA -> Table B, TableB -> TableA).

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -58,7 +58,7 @@ function mixinMigration(MySQL, mysql) {
 
     models = models || Object.keys(this._models);
 
-    async.each(models, function(model, done) {
+    async.eachSeries(models, function(model, done) {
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

ssh24:
@slnode test please

ssh24:
@DaGaMs Thank you for the patch! PR is ready to be landed.

PR #265, opened 2017-04-22
legacyDateTypeProcessing flag for datetime processing

darknos:
new connector option `legacyDateTypeProcessing` default `true`. set to `false` to use new behaviour as discussed in #257 and #149

### DIFF
--- a/test/date_types.test.js
+++ b/test/date_types.test.js
@@ -232,7 +232,7 @@ describe('MySQL DATE, DATETTIME, TIMESTAMP types on server with non local TZ (+0
 });
 
 var prepareModel = function(tz, done) {
-  db = getSchema({timezone: tz});
+  db = getSchema({timezone: tz, legacyDateTypeProcessing:false});
   DateModel = db.define('DateModel', {
     id: {type: Number, id: 1, generated: true},
     datetimeField: {type: Date, dataType: 'datetime', null: false},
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

ssh24:
@DaGaMs Do you ming adding some unit tests on this PR to take care of your end scenario?

test/date_types.test.js
```diff
@@ -232,7 +232,7 @@ describe('MySQL DATE, DATETTIME, TIMESTAMP types on server with non local TZ (+0
 });
 
 var prepareModel = function(tz, done) {
-  db = getSchema({timezone: tz});
+  db = getSchema({timezone: tz, legacyDateTypeProcessing:false});
```
superkhau: Why are we defaulting to false? We should be defaulting to true to preserve old behaviour and forcing users to turn on the flag for the new major/breaking behaviour.

superkhau:
@DaGaMs and @bbito Can you PTAL also? Would like to get more eyes on this.

bbito:
Hey @superkhau I wrote a long post yesterday discussing this and previous PR at https://github.com/strongloop/loopback-connector-mysql/issues/149#issuecomment-296501506 I also present 2 possible paths forward that I think are better options than https://github.com/strongloop/loopback-connector-mysql/pull/257 + https://github.com/strongloop/loopback-connector-mysql/pull/265 . Both of the options I present include the legacy support of this PR but with legacy behavior as default.

test/date_types.test.js
darknos: default was 'undefined' or true. these tests were on fixed date types processing. but it is not mater now, because #257 was reverted. :(

ssh24:
Closing this PR since https://github.com/strongloop/loopback-connector-mysql/pull/271 has been landed. Feel free to reopen if needed.

Issue #183, opened 2016-08-04
DATETIME: fractional seconds part


NTag:
Hello,

I have a model with a date field. In MySQL the type of the field is `DATETIME(3)` which means that it will store milliseconds (http://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html).
However, when I try to save a date which has milliseconds, in the database they disappear and I always get 0 milliseconds.

Do you support fractional seconds part?

Thank you very much,
Basile

jannyHou:
Didn't figure out a way to specify fractional seconds.
This is what I tried on local but doesn't work:
```js
  // Model json file
  "name": "Mytest",
  "properties": {
    "mydate": {
      "type": "date"
    },
    "fraction": {
      "type": "date"
    },
    "fraction4": {
      "type": "date",
      "mysql":
      {
        "columnName":"fraction4",
        "dataType":"DATETIME(4)"
      }
    }
  },
```
Data stored in mysql:
```js
//In explorer the instance is created as {"fraction4": "2016-12-06T19:42:38.5684"' }
// milliseconds disappeared and saved as '.0000'
fraction4:  '2016-12-06 19:42:38.0000'

```

nqd:
+1 to support ms supported

joostdebruijn:
I purposed a pull request for this. I'm awaiting feedback of the maintainers now.

jannyHou:
I will take a look asap

ssh24:
Should be fixed in https://github.com/strongloop/loopback-connector-mysql/pull/274

PR #270, opened 2017-04-25
Add support for fractional seconds.

joostdebruijn:
### Description
In MySQL 5.6 the support for saving fractional seconds for the DATETIME datatype was added. This pull request will add support for it to the connector.

I'm not sure about the best way to add test coverage for my changes to the current set of tests. If someone can help me with that, I will appreciate that. 😄

Note: MySQL supports a precision up to 6 digits, however this is not supported in JavaScript (it has a maximum of 3 digits). To avoid problems, I capped the maximum precision for the Date-datatype to three. If you agree with this limitation, I will add something in the README about that.

#### Related issues
#183 

### Checklist

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -886,11 +886,6 @@ function mixinMigration(MySQL, mysql) {
     var precision = 0;
 
     if (p.precision) {
-      // The JavaScript Date-object has a maximum precision of three for Dates, so we can't save it more precisely (although MySQL supports it).
-      if (p.precision > 3) {
-        p.precision = 3;
-      }
-
       precision = Number(p.precision);
       columnType += '(' + precision + ')';
     }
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

joostdebruijn:
I just realized the maximum precision of three doesn't matter for the connector when creating and maintaining the DATETIME column, so I removed it. When someone needs such precision for other matters without using JavaScript, it should be possible.

However, we may want to mention in the README that JavaScript is not able to parse a precision with more than three digits. It will cause data / precision loss when modifying such database entries via this connector.

joostdebruijn:
I noticed that in the last few days a lot of discussion about timezones took place. Some of the proposed changes (e.g. #271) will probably have influence on this pull request and the way fractional seconds could be supported for this connector.

My suggestion is to set this PR on hold until it is clear which of proposed timezone related changes will be merged in to the master. Then I will update this pull request or create a new one.

PR #273, opened 2017-04-27
Properties with mysql custom "columnName" don't get autoupdated

darknos:
### Description

#250 for 3.x

connect to #250 

### Checklist

- [+] New tests added or existing tests modified to cover all changes
- [+] Code conforms with the [style

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -176,7 +176,7 @@ function mixinMigration(MySQL, mysql) {
         });
       }
       if (found) {
-        actualize(colName, found);
+        actualize(propName, found);
       } else {
         sql.push('ADD COLUMN ' + self.client.escapeId(colName) + ' ' +
           self.buildColumnDefinition(model, propName));
@@ -186,7 +186,7 @@ function mixinMigration(MySQL, mysql) {
     function actualize(propName, oldSettings) {
       var newSettings = m.properties[propName];
       if (newSettings && changed(newSettings, oldSettings)) {
-        var pName = self.client.escapeId(propName);
+        var pName = self.columnEscaped(model, propName);
         sql.push('CHANGE COLUMN ' + pName + ' ' + pName + ' ' +
           self.buildColumnDefinition(model, propName));
       }
--- a/test/mysql.autoupdate.test.js
+++ b/test/mysql.autoupdate.test.js
@@ -528,7 +528,7 @@ describe('MySQL connector', function() {
     // second autoupdate call uses alter table
     verifyMysqlColumnNameAutoupdate(done);
   });
-  
+
   it('should update the nullable property of "first_name" to false', function(done) {
     // update the model "required" property
     var schema = {
--- a/test/mysql.autoupdate.test.js
+++ b/test/mysql.autoupdate.test.js
@@ -528,6 +528,49 @@ describe('MySQL connector', function() {
     // second autoupdate call uses alter table
     verifyMysqlColumnNameAutoupdate(done);
   });
+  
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

lib/migration.js
```diff
@@ -186,7 +186,7 @@ function mixinMigration(MySQL, mysql) {
     function actualize(propName, oldSettings) {
       var newSettings = m.properties[propName];
       if (newSettings && changed(newSettings, oldSettings)) {
-        var pName = self.client.escapeId(propName);
+        var pName = self.columnEscaped(model, propName);
```
ssh24: change this to `var pName = (newSettings.mysql && newSettings.mysql.columnName) || self.client.escapeId(propName);`

lib/migration.js
```diff
@@ -176,7 +176,7 @@ function mixinMigration(MySQL, mysql) {
         });
       }
       if (found) {
-        actualize(colName, found);
+        actualize(propName, found);
```
ssh24: Why do we need change this?

darknos: because we have a way how to find column name by property name but not vise versa.

lib/migration.js
darknos: but in this case columnName will be not escaped and we have possible issue with columnName like `from`.

darknos: I want to merge this PR asap, but why do you want to replace documented and widely used `self.columnEscaped` to that.

... ten lines above we use `var colName = expectedColNameForModel(propName, m);` for the same result. 

please, confirm the change and, please, explain the reason

ssh24: Right. But what I am not understanding is, if we specify the `columnName` explicitly to be something different than the property name, how would this work?

ssh24: Something like this?
```
var colName = newSettings.mysql && newSettings.mysql.columnName || self.client.escapeId(propName);
var pName = self.columnEscaped(model, colName);
```

darknos:
```javascript
var pName = self.client.escapeId(newSettings.mysql && newSettings.mysql.columnName || propName);
```
do the same as
```javascript
var pName = self.columnEscaped(propName);
```


```javascript
/**
 * Get the escaped column name for a given model property
 * @param {String} model The model name
 * @param {String} property The property name
 * @returns {String} The escaped column name
 */
SQLConnector.prototype.columnEscaped = function(model, property) {
  return this.escapeName(this.column(model, property));
};
```
```javascript
/**
 * Get the column name for the given model property. The column name can be
 * customized at the model property definition level as `column` or
 * `columnName`. For example,
 *
 * ```json
 * "name": {
 *   "type": "string",
 *   "mysql": {
 *     "column": "NAME"
 *   }
 * }
 * ```
 *
 * @param {String} model The model name
 * @param {String} property The property name
 * @returns {String} The column name
 */
SQLConnector.prototype.column = function(model, property) {
  var prop = this.getPropertyDefinition(model, property);
  var columnName;
  if (prop && prop[this.name]) {
    columnName = prop[this.name].column || prop[this.name].columnName;
    if (columnName) {
      // Explicit column name, return as-is
      return columnName;
    }
  }
  columnName = property;
  if (typeof this.dbName === 'function') {
    columnName = this.dbName(columnName);
  }
  return columnName;
};
```

ssh24:
@slnode ok to test

ssh24:
@slnode test please

ssh24:
@darknos Thank you for taking the initiative and submitting this patch.

PR #274, opened 2017-05-03
Adding support for fractional seconds

joostdebruijn:
### Description
In MySQL 5.6 the support for saving fractional seconds for the DATETIME datatype was added. This pull request will add support for it to the connector for the use of native Date-types. The DateString-type already supports fractional seconds.

Earlier this week I posted #270, however this a a replacement based on the changes in the 4.1-release of the connector.

#### Related issues
- #183 

### Checklist
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -723,7 +723,8 @@ function mixinMigration(MySQL, mysql) {
         dt = numericOptionsByType(p, dt);
         break;
       case 'Date':
-        dt = columnType(p, 'DATETIME'); // Currently doesn't need options.
+        dt = columnType(p, 'DATETIME');
+        dt = dateOptionsByType(p, dt);
         break;
       case 'Boolean':
         dt = 'TINYINT(1)';
@@ -890,6 +891,31 @@ function mixinMigration(MySQL, mysql) {
           break;
       }
     }
+    return columnType;
+  }
+
+  function dateOptionsByType(p, columnType) {
+    switch (columnType.toLowerCase()) {
+      default:
+      case 'date':
+        break;
+      case 'time':
+      case 'datetime':
+      case 'timestamp':
+        columnType = dateOptions(p, columnType);
+        break;
+    }
+    return columnType;
+  }
+
+  function dateOptions(p, columnType) {
+    var precision = 0;
+
+    if (p.precision) {
+      precision = Number(p.precision);
+      columnType += '(' + precision + ')';
+    }
+
     return columnType;
   }
 
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -412,7 +412,7 @@ MySQL.prototype.fromColumnValue = function(prop, val) {
         // MySQL allows, unless NO_ZERO_DATE is set, dummy date/time entries
         // new Date() will return Invalid Date for those, so we need to handle
         // those separate.
-        if (!val || val == '0000-00-00 00:00:00' || val == '0000-00-00') {
+        if (!val || /^0{4}(-00){2}( (00:){2}0{2}(\.0{1,6}){0,1}){0,1}$/.test(val)) {
           val = null;
         }
         break;
--- a/test/datetime.test.js
+++ b/test/datetime.test.js
@@ -17,6 +17,7 @@ describe('MySQL datetime handling', function() {
     married: Boolean,
     dob: {type: 'DateString'},
     createdAt: {type: Date, default: Date},
+    lastLogon: {type: Date, precision: 3, default: Date},
   };
 
   // Modifying the connection timezones mid-flight is a pain,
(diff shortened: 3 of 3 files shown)


slnode:

slnode:

slnode:

jannyHou:
@joostdebruijn Thank you, could you rebase your code? I will trigger CI test for you then.
And if you don't mind, can you give me permission to rebase your PR so you don't need to bother with the frequently checked in commits. permission guild: https://github.com/blog/2247-improving-collaboration-with-forks

joostdebruijn:
@jannyHou Thanks for having a first look at it! I gave you permissions to edit my fork. I'm not very familiar with doing rebases, so I will appreciate your help with that.

ssh24:
@joostdebruijn I have rebased the branch for you. Please sign the [Contributor License Agreement](https://cla.strongloop.com/agreements/strongloop/loopback-connector-mysql) before we can move on with this PR.

@slnode test please

joostdebruijn:
@ssh24 Done! I signed it already for my previous version of this PR (#270). Just to know: do I need to sign it for future PR's again?

lib/mysql.js
```diff
@@ -412,7 +412,7 @@ MySQL.prototype.fromColumnValue = function(prop, val) {
         // MySQL allows, unless NO_ZERO_DATE is set, dummy date/time entries
         // new Date() will return Invalid Date for those, so we need to handle
         // those separate.
-        if (!val || val == '0000-00-00 00:00:00' || val == '0000-00-00') {
+        if (!val || /^0{4}(-00){2}( (00:){2}0{2}(\.0{1,6}){0,1}){0,1}$/.test(val)) {
```
ssh24: Not sure what the end of the regex means
> {0,1}$/.test(val)

lib/migration.js
```diff
@@ -893,6 +894,31 @@ function mixinMigration(MySQL, mysql) {
...
   }
 
+  function dateOptionsByType(p, columnType) {
+    switch (columnType.toLowerCase()) {
+      default:
+      case 'date':
+        break;
+      case 'time':
+      case 'datetime':
+      case 'timestamp':
```
ssh24: Also should we only add precision fractional second for timestamp values? According to MySQL [documentation](https://dev.mysql.com/doc/refman/5.7/en/datetime.html)
> A DATETIME or TIMESTAMP value can include a trailing fractional seconds part in up to microseconds (6 digits) precision

lib/mysql.js
joostdebruijn: This regex checks if val is one of the following values:

0000-00-00
0000-00-00 00:00:00
0000-00-00 00:00:00.0
0000-00-00 00:00:00.00
0000-00-00 00:00:00.000
0000-00-00 00:00:00.0000
0000-00-00 00:00:00.00000
0000-00-00 00:00:00.000000

You can see it working over here: https://regex101.com/r/IkM0wc/2

lib/migration.js
joostdebruijn: I think I don't understand your comment. As the MySQL documentation said: [TIME](https://dev.mysql.com/doc/refman/5.7/en/datetime.html), DATETIME and TIMESTAMP types can have a precision. The DATE-type (of course) not.

This function checks the columntype and for columns with a type of time, datetime and timestamp and a precision is defined, the columnType will modified (see line 915), for example to DATETIME(3). This will allow to save values like 2017-05-07 12:34:56.789. When the column is of the type DATE, the function returns the colomnType without any modification.

lib/mysql.js
ssh24: Yes, I get what the regex does. I am confused about the end part of the regex.
> .test(val)

joostdebruijn: Ah, I get it! The `test()` method checks if the string specified in the function argument matches the regex. It will return true or false. See also [this article on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test).

ssh24: That makes much more sense now.

test/datetime.test.js
```diff
@@ -17,6 +17,7 @@ describe('MySQL datetime handling', function() {
     married: Boolean,
     dob: {type: 'DateString'},
     createdAt: {type: Date, default: Date},
+    lastLogon: {type: Date, precision: 3, default: Date},
```
jannyHou: hi @joostdebruijn from the login on line#900, a property with `date` type won't execute `dateOptions()` which appends precision config, can you take a look?

joostdebruijn: Are you sure? I ran the tests again and `dateOptions()` is hit twice for my specific test. The first time for property createdAt (skipping the if at [migration.js:914](https://github.com/joostdebruijn/loopback-connector-mysql/blob/c3a336c18dd9cebe1483d1fbe5b670a0496e26cf/lib/migration.js#L914)), resulting in a DATETIME. The second time for the property lastLogon (hitting the if at [migration.js:914](https://github.com/joostdebruijn/loopback-connector-mysql/blob/c3a336c18dd9cebe1483d1fbe5b670a0496e26cf/lib/migration.js#L914)) and resulting in a DATETIME(3).

Please note that the Loopback-type 'Date' maps on 'DATETIME' by default, see [documentation](https://loopback.io/doc/en/lb3/MySQL-connector.html#type-mappings) and [migration.js:725](https://github.com/joostdebruijn/loopback-connector-mysql/blob/c3a336c18dd9cebe1483d1fbe5b670a0496e26cf/lib/migration.js#L725). The method `dateOptions()` handles the columnType and not the type as in Loopback-type.

jannyHou: ah I see, it's executed after ` dt = columnType(p, 'DATETIME');`. Thanks.

joostdebruijn:
Thanks @ssh24 and @jannyHou for your questions and critical review!

Issue #269, opened 2017-04-25
GeoPoint stored as POINT(lat, lon) when it should be POINT(lon, lat)


tma-isbx:
From the code:

```
if (prop.type.name === 'GeoPoint') {
    return new ParameterizedSQL({
      sql: 'Point(?,?)',
      params: [val.lat, val.lng],
    });
  }
```

```
      case 'GeoPoint':
      case 'Point':
        val = {
          lat: val.x,
          lng: val.y,
        };
        break;
```

MySQL expects it to be POINT(X, Y) which would be POINT(lon, lat). Storing them transposed would cause confusion when interacting with SQL queries that interact with the POINT. Functions like ST_X(), ST_Y() would be confusing (at best). If MySQL were to support geography based functions that worked on a spheroid in the future, distance calculations would be incorrect.

`SELECT ST_X( POINT( 3, 2 ) ) AS x` returns 3.

ssh24:
@raymondfeng @bajtos Looks like a bug to me. Any reason why we specify the params as `[val.lat, val.lng]` rather than `[val.lng, val.lat]`? And of course need to change 
```
case 'Point':
        val = {
          lng: val.x,
          lat: val.y,
        };
```

bajtos:
IIRC, the last time I was looking into this issue, there was no generally accepted consensus on how to order lat+lng in an array. 

For example, Google Maps uses `(lat, long)` format - see https://developers.google.com/maps/documentation/javascript/reference#LatLng
```
map.setCenter(new google.maps.LatLng(-34, 151));
```

I believe we have modelled our GeoPoint API using Google's approach. Changing the order of lat/lng now would be a huge breaking change.

I think the right solution is to fix the MySQL connector to correctly load/save the GeoPoint type in the format that's used by MySQL.

cc @raymondfeng

b-admike:
@tma-isbx This should be fixed in the next released version of this connector. Thanks!

tma-isbx:
Sounds good!

Though this isn't a breaking change API-wise. It will be a pretty significant breaking change for users that already use GeoPoint with MySQL and have POINTs in their database.

We will lock our version of the connector and manually upgrade it in sync with a database migration. But I assume this could cause some surprises for other users.

bajtos:
> It will be a pretty significant breaking change for users that already use GeoPoint with MySQL and have POINTs in their database.

[...]

PR #282, opened 2017-05-29
switch long and lat vals for mysql

b-admike:
### Description

MySQL expects reverse order of latitude and longitude from the way we use it in LoopBack, so switch the order when saving and loading Point spatial type we use for Point/GeoPoint.

#### Related issues
connect to #269 

### Checklist

### DIFF
--- a/test/datatypes.test.js
+++ b/test/datatypes.test.js
@@ -7,7 +7,7 @@
 require('./init.js');
 var assert = require('assert');
 
-var db, BlobModel, EnumModel, ANIMAL_ENUM;
+var db, BlobModel, EnumModel, ANIMAL_ENUM, City;
 var mysqlVersion;
 
 describe('MySQL specific datatypes', function() {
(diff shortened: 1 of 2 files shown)


test/datatypes.test.js
```diff
@@ -86,6 +86,35 @@ describe('MySQL specific datatypes', function() {
...
+      },
+    };
+    var city2 = {
+      name: 'Saanich',
+      loc: {
+        lat: 48.459377,
+        lng: -123.378014,
+      },
+    };
+    city.create([city1, city2], function(err, res) {
```
jannyHou: nitpick: any reason why creating two instances here?

b-admike: It was for fun and to make sure where filter returns only one :-)

test/datatypes.test.js
```diff
@@ -110,6 +130,11 @@ function setup(done) {
     bin: {type: Buffer, dataType: 'blob', null: false},
     name: {type: String},
   });
+
+  city = db.define('City', {
```
jannyHou: usually we use upper case letter in a model name: `City`

bajtos:
I believe the new test was passing with the old implementation too, because the (wrong) order was used both for loading and saving. It would be nice to have a test that actually fails in the old implementation, I think it has to invoke custom SQL command using MySQL's Point features.

b-admike:
@bajtos Thank you; I will re-write the test.

Issue #263, opened 2017-04-20
Multi Column Index Throwing 500 error on duplicate creation


kodepareek:
Hi guys

I am on LB 3. 

I have `team` and `appUser` models which are linked by `hasManyThrough` via a `teamappUser` model. 

I created a unique together constraint on `teamId` and `appUserId` on in the `teamappUser` 

On testing for duplicate key creation (testing if unique together constraint was successful ) I get a 500 internal server error and server restart. 

I feel a 500 is completely unwarranted for what is expected behaviour on part of the app. This should be a 4XX error and should not cause the server to crash. 

package version 4.0.0

The Models are all very basic. 

Output on console. 

```
loopback:connector:mysql SQL: SELECT `id`,`ttl`,`created`,`userId`,`appUserId` FROM `AccessToken` WHERE `id`=? ORDER BY `id` LIMIT 1, params: ["l8vY592xJjF6SxrEawMvUPCzWNqZAZpwZNaccpgoANIdzzSmquAKaRFDLwPLrCDr"] +0ms
  loopback:connector:mysql Data:  +7ms [ RowDataPacket {
    id: 'l8vY592xJjF6SxrEawMvUPCzWNqZAZpwZNaccpgoANIdzzSmquAKaRFDLwPLrCDr',
    ttl: 1209600,
    created: 2017-04-19T07:33:07.000Z,
    userId: 1,
    appUserId: null } ]
  loopback:connector:mysql SQL: INSERT INTO `teamappUser`(`teamId`,`appUserId`) VALUES(?,?), params: [2,2] +50ms
  loopback:connector:mysql Error: {"code":"ER_DUP_ENTRY","errno":1062,"sqlState":"23000","index":0} +59ms
  loopback:connector:mysql Data:  +0ms undefined
Unhandled error for request POST /api/teamappUsers?access_token=l8vY592xJjF6SxrEawMvUPCzWNqZAZpwZNaccpgoANIdzzSmquAKaRFDLwPLrCDr: Error: ER_DUP_ENTRY: Duplicate entry '2-2' for key 'teamIdAppUserId'



```

kjdelisle:
@kodepareek Sorry about the super late reply, we're churning through quite a large backlog here!

Do you have an example repository that we can use to reproduce? I can try to recreate the problem myself, but having an example that you know demonstrates the issue would definitely speed this up.

kjdelisle:
@kodepareek An update:

[...]

PR #284, opened 2017-05-30
Add http status code translations for errors

kjdelisle:
### Description
Errors will now have their status codes set based on the error
message returned by the MySQL driver.
#### Related issues
#263 
### Checklist


lib/set-http-code.js
```diff
@@ -0,0 +1,30 @@
...
+  } else if (!(err instanceof Error)) {
+    err = new Error(err); // Sucks that we weren't given an error object...
+  }
+
+  // Convert errors into status codes here.
+  if (_.includes(err.message, 'ER_DUP_ENTRY')) {
+    err.statusCode = 422;
+  } else {
+    // If all else fails, call it a 500.
+    err.statusCode = 500;
```
ssh24: This implies if the error is not a duplicate entry then it will be always an internal server error?

b-admike: If I understand correctly, the idea to is to add more error types as applicable, but for now all else will be treated as 500.

kjdelisle:
Side note: I'm more than willing to expand on this PR immediately if there are any MySQL error codes that you know of that need different/better HTTP status codes to go with them!

lib/set-http-code.js
kjdelisle: ^ what he said

ssh24:
I think we can cherry pick some of the useful error codes from [MySQL Error codes](https://www.briandunning.com/error-codes/?source=MySQL) and handle them. Such as
- 1006	SQLSTATE: HY000 (ER_CANT_CREATE_DB) Can't create database '%s' (errno: %d)
- 1007	SQLSTATE: HY000 (ER_DB_CREATE_EXISTS) Can't create database '%s'; database exists
- 1008	SQLSTATE: HY000 (ER_DB_DROP_EXISTS) Can't drop database '%s'; database doesn't exist
- 1050	SQLSTATE: 42S01 (ER_TABLE_EXISTS_ERROR) Table '%s' already exists
- 1051	SQLSTATE: 42S02 (ER_BAD_TABLE_ERROR) Unknown table '%s'

It is not worth doing all of them, so we can have `500` (Internal server error) as the else part to handle everything else.

I have just chosen some of them, feel free to add more if you feel like they are applicable.

kjdelisle:
At first glance, those codes feel like they correspond to these HTTP Status Codes:
```
ER_CANT_CREATE_DB - 422 Unprocessable Entity
ER_DB_CREATE_EXISTS - 422 Unprocessable Entity
ER_DB_DROP_EXISTS - 404 Not Found
ER_TABLE_EXISTS_ERROR - 422 Unprocessable Entity
ER_BAD_TABLE_ERROR - 404 Not Found
```

Though, all of those suggestions look like they apply exclusively to automigrate/autoupdate for our purposes.

Are there any others for common CRUD operations that you've seen before?

ssh24:
> Are there any others for common CRUD operations that you've seen before?

Not that I recall of any specific ones.

lib/set-http-code.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+module.exports = function(err) {
+  if (!err) {
+    return;
+  } else if (!(err instanceof Error)) {
+    err = new Error(err); // Sucks that we weren't given an error object...
+  }
+  // Find error prefix
+  var msg = err.message;
+  var sqlError = msg.substring(0, msg.indexOf(':'));
```
ssh24: Do the mysql error come in the form of `SOME_ERROR:blahblahblah`?

kjdelisle: Yes, the errors bubbling up from the driver always come back in this form.

PR #286, opened 2017-06-01
switch long and lat for geopoint type

b-admike:
### Description

MySQL expects reverse order of latitude and longitude from the way we use it in LoopBack, so switch the order when saving and loading Point spatial type we use for Point/GeoPoint.

#### Related issues
connect to #269 
connected to https://github.com/strongloop-internal/scrum-apex/issues/239

### Checklist

### DIFF
--- a/test/datatypes.test.js
+++ b/test/datatypes.test.js
@@ -9,6 +9,7 @@ var assert = require('assert');
 var _ = require('lodash');
 
 var db, BlobModel, EnumModel, ANIMAL_ENUM, City, Account;
+
 var mysqlVersion;
 
 describe('MySQL specific datatypes', function() {
@@ -58,7 +59,7 @@ describe('MySQL specific datatypes', function() {
 
     it('create an instance', function(done) {
       Account.create(data, function(err, result) {
-        assert(!err);
+        if (err) return done(err);
         assert(result);
         assert(_.isEqual(data.length, result.length));
         assert(_.isEqual(data[0].amount, result[0].amount));
@@ -69,7 +70,7 @@ describe('MySQL specific datatypes', function() {
 
     it('find an instance', function(done) {
       Account.find(function(err, result) {
-        assert(!err);
+        if (err) return done(err);
         assert(result);
         assert(_.isEqual(data.length, result.length));
         assert(_.isEqual(data[0].amount, result[0].amount));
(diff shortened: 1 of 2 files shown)


kjdelisle:
@bajtos Are we okay to merge this? I've opened the 5.x-beta branch as a way to land breaking changes but keep them off master for the time being.

bajtos:
In the past, when we started to work on a new semver-major version, we created a new branch for the "latest" (4.x in this case) and kept "master" for the work on the upcoming major version. This way the git commit history remains linear. OTOH, when you fork off 5.x-beta branch and then merge it back to master later, then you will have two parallel lines in master history. There is also higher chance of getting merge conflicts, because by the time you merge 5.x-beta back into master, there may be other commits on master that are in conflict with 5.x-beta work.

Here is the description of the process we used for LoopBack 3.0: [link to gist](https://gist.github.com/bajtos/ede94ce84add33d0e391ee8e7d86c094), see also our internal https://github.com/strongloop-internal/scrum-loopback/issues/601

Last but not least: we need to consider implications for long-term support, as described in http://loopback.io/doc/en/contrib/Long-term-support.html

The important practical part is make sure that each branch we are actively publishing from (3.x, 4.x, etc.) has a unique npm tag, and the tag "latest" is used by the branch containing the stable GA version. For example, strong-remoting's master branch is publishing "3.x" versions tagged as "latest", while the 2.x branch is publishing "2.x" versions tagged as "lts".

This is the process we used in LoopBack, it worked pretty well for us IMO. Of course, feel free to choose a different process if it works better for your squad/team.

bajtos:
See https://github.com/npm/npm/issues/6778 for an explanation why we need different npm tags for different release lines.

b-admike:
>I believe the new test was passing with the old implementation too, because the (wrong) order was used both for loading and saving. It would be nice to have a test that actually fails in the old implementation, I think it has to invoke custom SQL command using MySQL's Point features.
@bajtos PTAL. I've added more checks which make use of MySQL's Point functions and would fail in the old way we saved and loaded the GeoPoint datatype.

test/datatypes.test.js
```diff
@@ -179,6 +180,34 @@ describe('MySQL specific datatypes', function() {
...
+    var city1 = {
+      name: 'North York',
+      loc: {
+        lat: 43.761539,
+        lng: -79.411079,
+      },
+    };
+    var xcor, ycor;
+    City.create(city1, function(err, res) {
+      assert.ok(!err);
```
bajtos: FWIW, we use the following pattern in other LoopBack projects - it produces more helpful failure messages:

```diff
- assert.ok(!err);
+ if (err) return done(err);
```

kjdelisle:
@b-admike Before this merges, don't forget to change the target branch based on your discussion with @bajtos

PR #288, opened 2017-06-05
Handle null values in inq properly

### LINKED ISSUE #221: model filter with inq: [null, 1] breaks
If I'm doing a model filter with something like...

Model.find({where: {id: {inq: [null, 1]}}}) the sql generated will build something that looks like this...

```
select <fields>
from model
where model.id in (DEFAULT, 1)
```

this is erroneous sql

Edit: this is happening on the latest release of this connector as well as previous releases

b-admike:
### Description
Fixes #221 
connect to connected to https://github.com/strongloop-internal/scrum-apex/issues/239

This PR is to handle `null` values in the `inq` operator for filters in MySQL properly. If the column is nullable, return `null`, otherwise try to coerce it to its type or return the string 'null'.
### Checklist

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -331,11 +331,25 @@ MySQL.prototype.getInsertedId = function(model, info) {
  * @returns {*} The escaped value of DB column
  */
 MySQL.prototype.toColumnValue = function(prop, val) {
-  if (val == null) {
-    if (prop.autoIncrement || prop.id) {
-      return new ParameterizedSQL('DEFAULT');
-    }
+  if (val === undefined && this.isNullable(prop)) {
     return null;
+  }
+  if (val === null) {
+    if (this.isNullable(prop)) {
+      return val;
+    } else {
+      try {
+        var castNull = prop.type(val);
+        if (prop.type === Object) {
+          return JSON.stringify(castNull);
+        }
+        return castNull;
+      } catch (err) {
+        //if we can't coerce null to a certain type,
+        //we just return it
+        return 'null';
+      }
+    }
   }
   if (!prop) {
     return val;
--- a/test/mysql.test.js
+++ b/test/mysql.test.js
@@ -6,7 +6,7 @@
 'use strict';
 var should = require('./init.js');
 
-var Post, PostWithStringId, PostWithUniqueTitle, db;
+var Post, PostWithStringId, PostWithUniqueTitle, PostWithNumId, db;
 
 // Mock up mongodb ObjectID
 function ObjectID(id) {
@@ -41,13 +41,22 @@ describe('mysql', function() {
       title: {type: String, length: 255, index: true},
       content: {type: String},
     });
-
+    PostWithNumId = db.define('PostWithNumId', {
+      id: {type: Number, id: true, null: false},
+      title: {type: String, length: 255, null: false, index: true},
+      content: {type: String, null: false},
+      buffProp: {type: Buffer, null: false},
+      objProp: {type: Object, null: false},
+      arrProp: {type: [Number], null: false},
+      dateProp: {type: Date, null: false},
+      pointProp: {type: 'GeoPoint', null: false},
+    });
     PostWithUniqueTitle = db.define('PostWithUniqueTitle', {
       title: {type: String, length: 255, index: {unique: true}},
       content: {type: String},
     });
 
-    db.automigrate(['PostWithDefaultId', 'PostWithStringId', 'PostWithUniqueTitle'], function(err) {
+    db.automigrate(['PostWithDefaultId', 'PostWithStringId', 'PostWithUniqueTitle', 'PostWithNumId'], function(err) {
       should.not.exist(err);
       done(err);
     });
@@ -56,9 +65,7 @@ describe('mysql', function() {
   beforeEach(function(done) {
     Post.destroyAll(function() {
       PostWithStringId.destroyAll(function() {
-        PostWithUniqueTitle.destroyAll(function() {
-          done();
-        });
+        PostWithUniqueTitle.destroyAll(done);
       });
     });
   });
@@ -421,6 +428,96 @@ describe('mysql', function() {
         should.not.exist(err);
         posts.should.have.property('length', 0);
         done();
+      });
+    });
+  });
+  context('null vals in different operators', function() {
+    var defaultPost = {
+      id: 3,
+      title: 'defTitle',
+      content: 'defContent',
+      buffProp: new Buffer('defBuffer'),
+      objProp: {defKey: 'defVal'},
+      arrProp: [0],
+      dateProp: new Date('2017-06-14'),
+      pointProp: {lng: 4.51515, lat: 57.2314},
+    };
+    beforeEach(function(done) {
+      PostWithNumId.destroyAll(done);
+    });
+    after(function(done) {
+      PostWithNumId.destroyAll(done);
+    });
+    it('should handle null in inq operator', function(done) {
+      defaultPost.id = 1;
+      defaultPost.title = 'Foo';
+      defaultPost.content = 'Bar';
+      PostWithNumId.create(defaultPost, function(err, post) {
+        should.not.exist(err);
+        post.id.should.equal(defaultPost.id);
+        PostWithNumId.find({where: {id: {inq: [null, 1]}}}, function(err, posts) {
+          should.not.exist(err);
+          posts.length.should.equal(1);
+          posts[0].title.should.equal('Foo');
+          posts[0].id.should.equal(1);
+          done();
+        });
+      });
+    });
+
+    it('should handle null in nin operator', function(done) {
+      defaultPost.id = 2;
+      defaultPost.title = 'Make';
+      defaultPost.content = 'Toyota';
+      PostWithNumId.create(defaultPost, function(err, post) {
+        should.not.exist(err);
+        post.id.should.equal(defaultPost.id);
+        PostWithNumId.find({where: {id: {nin: [null, 3]}}}, function(err, posts) {
+          should.not.exist(err);
+          posts.length.should.equal(1);
+          posts[0].content.should.equal('Toyota');
+          posts[0].id.should.equal(2);
+          done();
+        });
+      });
+    });
+
+    it('should handle null in neq operator', function(done) {
+      defaultPost.id = 3;
+      defaultPost.title = 'Model';
+      defaultPost.content = 'Corolla';
+      PostWithNumId.create(defaultPost, function(err, post) {
+        should.not.exist(err);
+        post.id.should.equal(defaultPost.id);
+        PostWithNumId.find({where: {id: {neq: null}}}, function(err, posts) {
+          should.not.exist(err);
+          posts.length.should.equal(1);
+          posts[0].content.should.equal('Corolla');
+          posts[0].id.should.equal(3);
+          done();
+        });
+      });
+    });
+
+    it('should handle null in nin op for different datatypes', function(done) {
+      PostWithNumId.create(defaultPost, function(err, post) {
+        should.not.exist(err);
+        post.id.should.equal(defaultPost.id);
+        PostWithNumId.find({where: {and: [
+        {id: {nin: [null]}},
+        {title: {nin: [null]}},
+        {content: {nin: [null]}},
+        {buffProp: {nin: [null]}},
+        {objProp: {nin: [null]}},
+        {arrProp: {nin: [null]}},
+        {dateProp: {nin: [null]}},
+        {pointProp: {nin: [null]}},
+        ]}}, function(err, posts) {
+          should.not.exist(err);
+          posts.length.should.equal(1);
+          posts[0].toObject().should.deepEqual(defaultPost);
+          done();
+        });
       });
     });
   });


test/mysql.test.js
```diff
@@ -425,6 +425,19 @@ describe('mysql', function() {
     });
   });
 
+  it.only('should handle null in inq operator', function(done) {
```
ssh24: Can we remove the `.only`? :P

lib/mysql.js
```diff
@@ -332,9 +332,6 @@ MySQL.prototype.getInsertedId = function(model, info) {
  */
 MySQL.prototype.toColumnValue = function(prop, val) {
   if (val == null) {
-    if (prop.autoIncrement || prop.id) {
-      return new ParameterizedSQL('DEFAULT');
```
ssh24: If we remove this line then it won't return the `DEFAULT(columnName)`, AFAIK

test/mysql.test.js
```diff
@@ -425,6 +425,45 @@ describe('mysql', function() {
     });
   });
 
+  it('should handle null in inq operator', function(done) {
+    Post.create({title: 'Foo', content: 'Bar'}, function(err, post) {
+      should.not.exist(err);
+      Post.find({where: {id: {inq: [null, 1]}}}, function(err, posts) {
```
jannyHou: nitpick: maybe we can turn off `forceId`, then create the instance with `id`, to make sure the idValue we put in filter is the one we created.

Same for the following two use cases.

b-admike:
Here is the sql it generates to test the code:
```javascript
 loopback:connector:mysql SQL: SELECT `id`,`title`,`content`,`buffProp`,`objProp`,`arrProp`,`dateProp`,`pointProp` FROM `PostWithNumId` WHERE (`id` IN (?)) AND (`title` IN (?)) AND (`content` IN (?)) AND (`buffProp` IN (?)) AND (`objProp` IN (?)) AND (`arrProp` IN (?)) AND (`dateProp` IN (?)) AND (`pointProp` IN (?)) ORDER BY `id`, params: [0,null,null,null,"{}",null,"Thu Jun 08 2017 17:01:54 GMT-0400 (EDT)",null]
```

lib/mysql.js
kjdelisle: I'm not certain what the best practice here would be; do we want to have some sort of `DEFAULT` setting, or give people the option to set a default value if they configure it to work this way?

lib/mysql.js
```diff
@@ -332,10 +332,21 @@ MySQL.prototype.getInsertedId = function(model, info) {
...
+    if (this.isNullable(prop)) {
+      return null;
+    } else {
+      try {
+        var castNull = prop.type(null);
+        if (prop.type == Object) {
```
kjdelisle: Double equals!? :-1:

test/mysql.test.js
```diff
@@ -57,7 +66,9 @@ describe('mysql', function() {
     Post.destroyAll(function() {
       PostWithStringId.destroyAll(function() {
         PostWithUniqueTitle.destroyAll(function() {
-          done();
+          PostWithNumId.destroyAll(function() {
+            done();
```
kjdelisle: `PostWithNumId.destroyAll(done)` would do the same thing. If you're not doing extra work before the `done` call, might as well save some space on the call stack.

test/mysql.test.js
```diff
@@ -425,6 +436,77 @@ describe('mysql', function() {
...
+        {id: {inq: [null]}},
+        {title: {inq: [null]}},
+        {content: {inq: [null]}},
+        {buffProp: {inq: [null]}},
+        {objProp: {inq: [null]}},
+        {arrProp: {inq: [null]}},
+        {dateProp: {inq: [null]}},
+        {pointProp: {inq: [null]}},
+      ]}}, function(err, posts) {
+        //the main check here is that we produce a valid sql
```
kjdelisle: I don't think this is true.
I understand that syntactically invalid SQL would throw an error, but you could easily have a logic error that produced _semantically_ invalid SQL that does something, just not the _right_ thing.

test/mysql.test.js
```diff
@@ -425,6 +436,77 @@ describe('mysql', function() {
...
+        should.not.exist(err);
+        posts.length.should.equal(0);
```
kjdelisle: Based on my previous comment, I would create a situation where only a syntactically _and_ semantically valid SQL statement would return the correct number of rows, and then I would assert that those values were returned in correct quantity and form.

test/mysql.test.js
```diff
@@ -41,13 +41,22 @@ describe('mysql', function() {
       title: {type: String, length: 255, index: true},
       content: {type: String},
     });
-
+    PostWithNumId = db.define('PostWithNumId', {
+      id: {type: Number, id: true},
```
jannyHou: I might wrong, `this.isNullable(prop)` still returns false if we don't specify `null:true` right?
I remember yesterday you mentioned MYSQL allows id to be null, so I am not sure do you want to test id nullable here or id not nullable..

test/mysql.test.js
```diff
@@ -425,6 +434,82 @@ describe('mysql', function() {
     });
   });
 
+  it('should handle null in inq operator', function(done) {
+    PostWithNumId.create({id: 1, title: 'Foo', content: 'Bar'}, function(err, post) {
+      should.not.exist(err);
```
ssh24: Can we check for existence of `post` here?

test/mysql.test.js
```diff
@@ -425,6 +434,82 @@ describe('mysql', function() {
...
+        posts[0].title.should.equal('Foo');
+        posts[0].id.should.equal(1);
+        done();
+      });
+    });
+  });
+
+  it('should handle null in nin operator', function(done) {
+    PostWithNumId.create({id: 2, title: 'Make', content: 'Toyota'}, function(err, post) {
+      should.not.exist(err);
```
ssh24: Same. Check existence of `post`.

test/mysql.test.js
```diff
@@ -425,6 +434,82 @@ describe('mysql', function() {
...
+        posts[0].content.should.equal('Toyota');
+        posts[0].id.should.equal(2);
+        done();
+      });
+    });
+  });
+
+  it('should handle null in neq operator', function(done) {
+    PostWithNumId.create({id: 3, title: 'Model', content: 'Corolla'}, function(err, post) {
+      should.not.exist(err);
```
ssh24: Same here. Check existence for `post`.

test/mysql.test.js
```diff
@@ -425,6 +434,82 @@ describe('mysql', function() {
...
+      content: 'defContent',
+      buffProp: new Buffer('defBuffer'),
+      objProp: {defKey: 'defVal'},
+      arrProp: [0],
+      dateProp: new Date('11/12/2017'),
+      pointProp: {lng: 4.51515, lat: 57.2314},
+    };
+
+    PostWithNumId.create(defaultPost, function(err, post) {
+      should.not.exist(err);
```
ssh24: Same here. Check existence for `post`.

test/mysql.test.js
```diff
@@ -57,7 +66,7 @@ describe('mysql', function() {
     Post.destroyAll(function() {
       PostWithStringId.destroyAll(function() {
         PostWithUniqueTitle.destroyAll(function() {
-          done();
+          PostWithNumId.destroyAll(done);
```
ssh24: Won't need this if we are using `context`

test/mysql.test.js
```diff
@@ -753,7 +838,9 @@ describe('mysql', function() {
   after(function(done) {
     Post.destroyAll(function() {
       PostWithStringId.destroyAll(function() {
-        PostWithUniqueTitle.destroyAll(done);
+        PostWithUniqueTitle.destroyAll(function() {
+          PostWithNumId.destroyAll(done);
```
ssh24: Won't need this if you are using `context`.

test/mysql.test.js
```diff
@@ -424,6 +431,101 @@ describe('mysql', function() {
...
+        {pointProp: {nin: [null]}},
+        ]}}, function(err, posts) {
+          should.not.exist(err);
+          posts.length.should.equal(1);
+          posts[0].id.should.equal(defaultPost.id);
+          posts[0].title.should.equal(defaultPost.title);
+          posts[0].content.should.equal(defaultPost.content);
+          posts[0].buffProp.should.deepEqual(defaultPost.buffProp);
+          posts[0].objProp.should.deepEqual(defaultPost.objProp);
+          posts[0].pointProp.should.deepEqual(defaultPost.pointProp);
```
Amir-61: Doesn't the following work?
```
posts[0].should.eql(defaultPost);
```
If this works it is much simpler and you can get rid of a bunch of lines.

ssh24: No. The whole point of the test case it is to check for each of the properties.

jannyHou:
@b-admike Could you take a look of failures? Your change is reasonable to me, but not sure why tests fail..

lib/mysql.js
```diff
@@ -331,12 +331,26 @@ MySQL.prototype.getInsertedId = function(model, info) {
...
+      try {
+        var castNull = prop.type(val);
+        if (prop.type === Object) {
+          return JSON.stringify(castNull);
+        }
+        return castNull;
+      } catch (err) {
+        //if we can't coerce null to a certain type,
+        //we just return it
+        return 'null';
```
kjdelisle: If this is specifically to make a query string where the value is null, and it fails our `val === null` and `.isNullable` checks, why not always set the string to null instead of attempting to stringify something?

test/mysql.test.js
kjdelisle: If you're using the should.js `deepEqual` function, then it should be doing this for you, right?
Should.js uses its own version of Node.js core's deepEqual function
https://nodejs.org/api/assert.html#assert_assert_deepequal_actual_expected_message

It should traverse the inner structure and check the values for logical equality.
So, I'd say you can just write `posts[0].should.deepEqual(defaultPost)`;

kjdelisle: If you're concerned the meaning of the test will not be easily conveyed to the reader, then you could loop through the properties and assert equality programmatically, since both objects should share the same property keys.
```
for (var key in defaultPost) {
  posts[0][key].should.equal(defaultPost[key]);
}
```

lib/mysql.js
b-admike: Because of the following:

test/mysql.test.js
b-admike: @Amir-61 @kjdelisle I did that at first, but `posts[0]` has way more keys in it than `defaultPost`.
```javascript
 1) mysql null vals in different operators should handle null in nin op for different datatypes:

      Uncaught AssertionError: expected ModelConstructor {
  __cachedRelations: Object {},
  __data: Object {
    arrProp: List [ 0 ],
    buffProp: Buffer [ 64, 65, 66, 42, 75, 66, 66, 65, 72 ],
    content: 'defContent',
    dateProp: 2017-06-14 11:02:13.000 -0400,
    id: 3,
    objProp: Object { defKey: 'defVal' },
    pointProp: Object { lat: 57.2314, lng: 4.51515 },
    title: 'defTitle'
  },
  __dataSource: undefined,
  __persisted: true,
  __strict: true,
  __unknownProperties: Array []
} to equal Object {
  arrProp: Array [ 0 ],
  buffProp: Buffer [ 64, 65, 66, 42, 75, 66, 66, 65, 72 ],
  content: 'defContent',
  dateProp: 2017-06-14 11:02:12.984 -0400,
  id: 3,
  objProp: Object { defKey: 'defVal' },
  pointProp: Object { lat: 57.2314, lng: 4.51515 },
  title: 'defTitle'
}
      + expected - actual

       {
      -  "__cachedRelations": {}
      -  "__data": {
      -    "arrProp": [
      -      0
      -    ]
      -    "buffProp": [Buffer: [
      -      100
      -      101
      -      102
      -      66
      -      117
      -      102
      -      102
      -      101
      -      114
      -    ]]
      -    "content": "defContent"
      -    "dateProp": [Date: 2017-06-14T15:02:13.000Z]
      -    "id": 3
      -    "objProp": {
      -      "defKey": "defVal"
      -    }
      -    "pointProp": {
      -      "lat": 57.2314
      -      "lng": 4.51515
      -    }
      -    "title": "defTitle"
      +  "arrProp": [
      +    0
      +  ]
      +  "buffProp": [Buffer: [
      +    100
      +    101
      +    102
      +    66
      +    117
      +    102
      +    102
      +    101
      +    114
      +  ]]
      +  "content": "defContent"
      +  "dateProp": [Date: 2017-06-14T15:02:12.984Z]
      +  "id": 3
      +  "objProp": {
      +    "defKey": "defVal"
         }
      -  "__dataSource": [undefined]
      -  "__persisted": true
      -  "__strict": true
      -  "__unknownProperties": []
      +  "pointProp": {
      +    "lat": 57.2314
      +    "lng": 4.51515
      +  }
      +  "title": "defTitle"
       }
```

jannyHou: @b-admike maybe you can try `post[0].toObject()` or check [`hasOwnProperty`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty)

ssh24: @b-admike You need to assert it against `posts[0].__data`.

ssh24: +1 for `.toObject()`

b-admike: Thank you both. Both work fine, and I've applied one of them.

kjdelisle:
LGTM

b-admike:
@slnode test please

Issue #292, opened 2017-06-12
Allow regexp operation with case sensitvity


ssh24:
# Description/Steps to reproduce
MySQL's regexp query is not case sensitive. More info [here](https://dev.mysql.com/doc/refman/5.7/en/regexp.html). This feature allows regexp operation with case sensitivity.

**Note:** MySQL uses `BINARY` flag to allow regexp query with case sensitivity because the default regexp query is not case sensitive.

# Expected result
This code snippet will return `My Post`, `my post` and `my post 2`. Upon the completion of the feature, it shall only return `My Post` because of the regular expersion is: `'/^M.*$/'`.
```js
  mysqlDs.automigrate(function(err) {
    if (err) throw err;
    Post.create([{title: 'My Post', content: 'Hello'},
    {title: 'my post', content: 'hello'},
    {title: 'my post 2', content: 'hello'},
    {title: 'cy post', content: 'hello'}],
    function(err, result) {
      if (err) throw err;
      console.log('\nCreate: ' + JSON.stringify(result));
      Post.find({where: {title: {regexp: '/^M.*$/'}}},
      function(err, result) {
        if (err) throw err;
        console.log('\nFind: ' + JSON.stringify(result));
      });
    });
  });
```

PR #293, opened 2017-06-12
Add case sensitive regex

ssh24:
### Description
Allow case sensitivity on a regex operation.

connect to https://github.com/strongloop/loopback-connector-mysql/issues/292

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -529,16 +529,23 @@ MySQL.prototype.ping = function(cb) {
 MySQL.prototype.buildExpression = function(columnName, operator, operatorValue,
     propertyDefinition) {
   if (operator === 'regexp') {
+    var clause = columnName + ' REGEXP ?';
+    // By default, MySQL regexp is not case sensitive. (https://dev.mysql.com/doc/refman/5.7/en/regexp.html)
+    // To allow case sensitive regexp query, it has to be binded to a `BINARY` type.
+    // If ignore case is not specified, search it as case sensitive.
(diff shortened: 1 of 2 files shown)


lib/mysql.js
```diff
@@ -529,16 +529,23 @@ MySQL.prototype.ping = function(cb) {
 MySQL.prototype.buildExpression = function(columnName, operator, operatorValue,
     propertyDefinition) {
   if (operator === 'regexp') {
+    var clause = columnName + ' REGEXP ?';
+    // By default, MySQL regexp is not case sensitive. (https://dev.mysql.com/doc/refman/5.7/en/regexp.html)
+    // To allow, case sensitive regexp query, it has to be binded to a `BINARY` type.
```
b-admike: nitpick: `To allow case sensitive regexp query,`

test/mysql.test.js
```diff
@@ -633,6 +633,15 @@ describe('mysql', function() {
               });
             });
       });
+
+      it('filter with case sensitive regex string', function(done) {
```
b-admike: Suggestion: can we make the regex match something for string and regex object? i.e. have results

ssh24: I see what you mean. I also thought about doing that but those tests are already covered. Look:
https://github.com/strongloop/loopback-connector-mysql/blob/master/test/mysql.test.js#L586
https://github.com/strongloop/loopback-connector-mysql/blob/master/test/mysql.test.js#L703

PR #297, opened 2017-06-14
Create database on the docker script

ssh24:
### Description
Before we would not create the database `testdb` or the user input parameter on the docker script. This was mainly because mysql has a flag "createDatabase" set to true if the database does not exist.

Add the ability to create the database as a part of the spawning container process because currently if someone needs to spawn a mysql instance for other purposes (not to run the test suite), they will have to run the tests in order for that database to be created.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -64,7 +64,7 @@ docker cp ./test/schema.sql $MYSQL_CONTAINER:/home/ > /dev/null 2>&1
 printf "\n${CYAN}Container is up and running.${PLAIN}\n"
 
 ## export the schema to the mysql database
-printf "\n${RED}>> Exporting schema to database${PLAIN} ${GREEN}...${PLAIN}\n"
+printf "\n${RED}>> Exporting default schema${PLAIN} ${GREEN}...${PLAIN}\n"
 
 ## command to export schema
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -96,6 +96,17 @@ if [ "$TIMEOUT" -le 0 ]; then
 fi
 printf "\n${CYAN}Successfully exported schema to database.${PLAIN}\n"
 
+## create the database
+printf "\n${RED}>> Creating the database${PLAIN} ${GREEN}...${PLAIN}"
+docker exec -it $MYSQL_CONTAINER /bin/sh -c "mysql -u$USER -p$PASSWORD -e \"DROP DATABASE IF EXISTS $DATABASE\"" > /dev/null 2>&1
```
jannyHou: is there a space between `-u` and `$USER`? same for password

ssh24: Not for mysql. The syntax for mysql command line is `mysql -u<USER> -p<PASS>`

ssh24:
@slnode test please

setup.sh
jannyHou: ah never mind, I checked the mysql command but not docker command. LGTM

Issue #85, opened 2015-03-11
Can't define a bigint(20) auto increment PK


didikeke:
I am trying to create a bigint(20) auto increment PK. But it seems that does not support auto increment if it is not default PK dataType. 

Model properties:

```
 "properties": {
    "id": {
      "type": "number",
      "id": true,
      "mysql": {
        "dataType": "bigint"
      }
    }
  }
```

Searched codes, both 
https://github.com/strongloop/loopback-connector-mysql/blob/master/lib/mysql.js#L962-L965
and 
https://github.com/strongloop/loopback-connector-mysql/blob/master/lib/mysql.js#L1042-L1049

do not have a field to output 'AUTO_INCREMENT' if it is customized type for the PK.

Searching the doc: http://docs.strongloop.com/display/public/LB/Model+definition+JSON+file#ModeldefinitionJSONfile-Generalpropertyproperties

There is no such field to add  'AUTO_INCREMENT'.

Any idea?

tngan:
@didikeke Did you resolve this issue ?

didikeke:
@tngan I have created sql files to import instead of using autoupdate for a long time.

Tested just now,  bigint(20) auto increase of autoupdate still not working. 

by the way, if you create bigint PK **without** auto increase, you can use this code.

`
"id": {
  "type": "number",
  "id": true,
  "dataType": "bigint"
}
`

tlloydukdev:
Seems to still be the case, if you set generated: true then the dataType changes back to int(11) rather than bigint(20).

Any workarounds for this or plans to resolve ?

ssh24:
After doing some research, I can confirm it is a bug.

To have a `AUTO_INCREMENT PK`, in loopback model definition, specify the **generated: true** flag. However, it always uses `int(11)` as the datatype even when the user specifies the specific datatype.

PR #300, opened 2017-06-16
Honour user specific pk datatype

ssh24:
### Description
Currently, the primary key in mysql connector always assumes the datatype to be of `int(11)` if the **generated** flag is true. This does not honour user specific datatypes if one decides to specify it on the model definition.

connect to https://github.com/strongloop/loopback-connector-mysql/issues/85

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -582,8 +582,8 @@ function mixinMigration(MySQL, mysql) {
         var idName = this.idName(model);
         var idProp = this.getModelDefinition(model).properties[idName];
         if (idProp.generated) {
-          sql.push(self.columnEscaped(model, idName) +
-            ' INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY');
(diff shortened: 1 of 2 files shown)


lib/migration.js
```diff
@@ -582,8 +582,8 @@ function mixinMigration(MySQL, mysql) {
         var idName = this.idName(model);
         var idProp = this.getModelDefinition(model).properties[idName];
         if (idProp.generated) {
-          sql.push(self.columnEscaped(model, idName) +
-            ' INT(11) NOT NULL AUTO_INCREMENT PRIMARY KEY');
+          sql.push(self.columnEscaped(model, idName) + ' ' +
+          self.buildColumnDefinition(model, idName) + ' AUTO_INCREMENT PRIMARY KEY');
```
0candy: This line needs to be indented.

0candy:
Needs rebasing.

Issue #290, opened 2017-06-07
Add migration guide for MySQL Connector 5.x release


b-admike:
Let our users know what's changed in loopback-connector-mysql@5.x.

Acceptance Criteria:
- Tag PR with `breaking-change` label 
- Describe the migration guide in README
- Test the upgrade path according to migration guide

PR #302, opened 2017-06-16
Document GeoPoint migration

b-admike:
### Description
Connect to https://github.com/strongloop/loopback-connector-mysql/issues/290

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,7 +442,63 @@ module.exports = function(app) {
   });
 };
 ```
+#### Breaking Changes with GeoPoint since 5.x
+Prior to `loopback-connector-mysql@5.x`, MySQL connector was saving and loading GeoPoint properties from the MySQL database in reverse.
+MySQL expects values to be POINT(X, Y) or POINT(lng, lat), but the connector was saving them in the opposite order(i.e. POINT(lat,lng)).
+If you have an application with a model that has a GeoPoint property using previous versions of this connector, you can migrate your models
+using the following programmatic approach:
+**NOTE** Please back up the database tables that have your application data before performing any of the steps.
+1. Create a boot script under `server/boot/` directory with the following:
+```js
+'use strict';
+module.exports = function(app) {
+  function findAndUpdate() {
+    var teashop = app.models.teashop;
+    //find all instances of the model we'd like to migrate
+    teashop.find({}, function(err, teashops) {
+      teashops.forEach(function(teashopInstance) {
+        //what we fetch back from the db is wrong, so need to revert it here
+        var newLocation = {lng: teashopInstance.location.lat, lat: teashopInstance.location.lng};
+        //only update the GeoPoint property for the model
+        teashopInstance.updateAttribute('location', newLocation, function(err, inst) {
+          if (err)
+            console.log('update attribute failed ', err);
+          else
+        console.log('updateAttribute successful');
+        });
+      });
+    });
+  }
 
+  findAndUpdate();
+};
+```
+2. Run the boot script by simply running your application or `node .`
+
+For the above example, the model definition is as follows:
+```json
+{
+  "name": "teashop",
+  "base": "PersistedModel",
+  "idInjection": true,
+  "options": {
+    "validateUpsert": true
+  },
+  "properties": {
+    "name": {
+      "type": "string",
+      "default": "storename"
+    },
+    "location": {
+      "type": "geopoint"
+    }
+  },
+  "validations": [],
+  "relations": {},
+  "acls": [],
+  "methods": {}
+}
+```
 
 ## Running tests
 


README.md
```diff
@@ -442,29 +442,82 @@ module.exports = function(app) {
   });
 };
 ```
+#### Migrating models with GeoPoint types before loopback-connector-mysql@5.x
+Previous to `loopback-connector-mysql@5.x`, MySQL connector was saving and loading GeoPoint properties from the MySQL database in reverse.
+MySQL expects values to be POINT(X, Y) or POINT(lon, lat), but the connector was doing the opposite. If you have an application with a model
```
0candy: Should probably stick to using `lng` instead of `lon`?

ssh24: maybe add `but the connector was doing the opposite (i.e POINT(lat, lon))`

README.md
```diff
@@ -442,29 +442,82 @@ module.exports = function(app) {
...
+  },
+  "validations": [],
+  "relations": {},
+  "acls": [],
+  "methods": {}
+}
+```
 
-## Running tests
+## running tests
```
dhmlau: I think we should keep the original format of capitalization. e.g. `Running tests` instead of `running tests`.  
The similar changes below don't seem to be necessary to me.

ssh24: What is the need of changing all of these docs?

README.md
```diff
@@ -442,29 +442,82 @@ module.exports = function(app) {
...
 ```bash
-SET MYSQL_HOST=<HOST> SET MYSQL_PORT=<PORT> SET MYSQL_USER=<USER> SET MYSQL_PASSWORD=<PASSWORD> SET MYSQL_DATABASE=<DATABASE> SET CI=true npm test
+set mysql_host=<host> set mysql_port=<port> set mysql_user=<user> set mysql_password=<password> set mysql_database=<database> set ci=true npm test
 ```
 
-### Docker
-If you do not have a local MySQL instance, you can also run the test suite with very minimal requirements.
-- Assuming you have [Docker](https://docs.docker.com/engine/installation/) installed, run the following script which would spawn a MySQL instance on your local:
+### docker
+if you do not have a local mysql instance, you can also run the test suite with very minimal requirements.
```
0candy: Why did we changes the cases?

README.md
```diff
@@ -442,29 +442,82 @@ module.exports = function(app) {
   });
 };
 ```
+#### Migrating models with GeoPoint types before loopback-connector-mysql@5.x
```
dhmlau: Suggestion: Making the title more prominent because it is a breaking change.  Perhaps something like "Breaking changes with GeoPoint since 5.x "

README.md
```diff
@@ -442,29 +442,82 @@ module.exports = function(app) {
...
+Previous to `loopback-connector-mysql@5.x`, MySQL connector was saving and loading GeoPoint properties from the MySQL database in reverse.
+MySQL expects values to be POINT(X, Y) or POINT(lon, lat), but the connector was doing the opposite. If you have an application with a model
+that has a GeoPoint property using previous versions of this connector, you can migrate your models using the following programmatic approach:
+**NOTE** Please back up the database tables that have your application data before performing any of the steps.
+1. Create a boot script under `server/boot/` directory with the following:
```
dhmlau: there's only one step.  we can remove "1."

ssh24: There is a second step. To run the bootscript :P

README.md
b-admike: I like that; added.

README.md
b-admike: No idea how that happened. Fixed.

README.md
b-admike: Agreed; I've removed it.

README.md
b-admike: No idea how that happened. Fixed.

README.md
b-admike: Good catch both :). I've added a second step, but can remove it if it's too trivial.

README.md
b-admike: No idea how that happened. Fixed.

README.md
```diff
@@ -442,7 +442,62 @@ module.exports = function(app) {
   });
 };
 ```
+#### Breaking Changes with GeoPoint since 5.x
+Previous to `loopback-connector-mysql@5.x`, MySQL connector was saving and loading GeoPoint properties from the MySQL database in reverse.
+MySQL expects values to be POINT(lon, lat), but the connector was doing the opposite(i.e. POINT(lat,lon)). If you have an application with a model
```
0candy: Now there are 2 of them `lon`?

README.md
```diff
@@ -442,7 +442,62 @@ module.exports = function(app) {
   });
 };
 ```
+#### Breaking Changes with GeoPoint since 5.x
+Previous to `loopback-connector-mysql@5.x`, MySQL connector was saving and loading GeoPoint properties from the MySQL database in reverse.
```
dhmlau: sorry.. not sure if that's more accurately in english.. should it be "Prior to"?

dhmlau: nevermind. just google. i think this is good too.

b-admike: I like `Prior to`, changed.

b-admike:
@slnode test please

Issue #294, opened 2017-06-13
Add docs on foreign key support for automigrate/autoupdate


haotangio:
Current the `automigrate` and `autoupdate` methods do not create foreign keys in the database. Do we have any plan to proceed with it?

ssh24:
@haotangio It does create foreign keys in the database when specified. Look at https://github.com/strongloop/loopback-connector-mysql/blob/master/lib/migration.js#L68

haotangio:
This is my `automigrate.js` script applying to https://github.com/strongloop/loopback-example-relations repo:
```
var path = require('path');

var app = require(path.resolve(__dirname, '../server/server'));
var ds = app.datasources.db;
ds.autoupdate(null, function(err) {
  if (err) throw err;
  console.log('Finished migration');
  ds.disconnect();
});
```

But there is still no foreign key created, do I need to add something to the code?
![upload](https://user-images.githubusercontent.com/8952752/27165557-a45fde2c-51cf-11e7-9edc-484cb0b3ef27.PNG)

haotangio:
After a deep reverse engineering for the source coe, I figured out the way to achieve the purpose:

```
{
  "name": "Book",
  "base": "PersistedModel",
  "idInjection": true,
  "properties": {
    "name": {
      "type": "string"
    }
  },
  "validations": [],
  "relations": {
    "author": {
      "type": "belongsTo",
      "model": "Author",
      "foreignKey": "authorId",
      "primaryKey": "id"
    }
  },
  "acls": [],
  "methods": {},
  "foreignKeys": {
    "authorId": {
      "name": "authorId",
      "foreignKey": "authorId",
      "entityKey": "id",
      "entity": "Author"
    }
  }
}
```

But it's very very strange as why loopback didn't put this information in any documents.
Or is there any document that I missed?

ssh24:
@haotangio Good catch with the docs. I will add a quick PR for the fix.

ssh24:
Docs and example landed on MySQL specific readme and should be accessible in [loopback.io](http://loopback.io/doc/) with the next release of docs.

haotangio:
Thank you.
Btw, is there any `secret options` which are helpful but not documented there? :)

ssh24:
Yup. Look in this [doc](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html) for all the options available globally on each connector module.

dionisisZyg:
[...]

PR #303, opened 2017-06-20
Fix foreign key example to avoid constraint errors

ssh24:
### Description
Fix the foreign key example to avoid both way foreign key creation. It causes foreign key constraint failures since one of the models is created before the other one and contains a fk to the other model.

connect to https://github.com/strongloop/loopback-connector-mysql/issues/294

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,21 +363,26 @@ Example:
 {
   "name": "Book",
   "base": "PersistedModel",
-  "idInjection": true,
+  "idInjection": false,
   "properties": {
+    "bId": {
+      "type": "number",
+      "id": true,
+      "required": true
+    },
     "name": {
       "type": "string"
-    }, "isbn": {
-      "type": "string"
     },
+    "isbn": {
+      "type": "string"
+    }
   },
   "validations": [],
   "relations": {
     "author": {
       "type": "belongsTo",
       "model": "Author",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -434,11 +430,16 @@ module.exports = function(app) {
   var Book = app.models.Book;
   var Author = app.models.Author;
 
-  mysqlDs.automigrate(function(err) {
+  // first autoupdate the `Author` model to avoid foreign key constraint failure
+  mysqlDs.autoupdate('Author', function(err) {
     if (err) throw err;
+    console.log('\nAutoupdated table `Author`.');
```
kjdelisle: Why do we have the line break here? Just for formatting?

README.md
```diff
@@ -434,11 +430,16 @@ module.exports = function(app) {
...
 
-    // at this point the database tables `Book` and `Author`
-    // should have the foreign keys (`bookId` and `authorId`) integrated
+    mysqlDs.autoupdate('Book', function(err) {
+      if (err) throw err;
+      console.log('\nAutoupdated table `Book`.');
```
kjdelisle: ^ ditto ^

README.md
ssh24: Yes, just wanted to format it.

PR #308, opened 2017-07-14
Set isNewInstance to undefined for after save hook on replaceOrCreate and updateOrCreate

### LINKED ISSUE #106: ctx.isNewInstance not supported for 'before|after save' operation hooks on instance.save()?
I am seeing ctx.isNewInstance = undefined/true on an existing instance save. My script just loads all Person entities, then loops over and updates a single attrib value. Works fine when I use `.updateAttribute()` but it should work for `.save()` since I'm calling a function on an instance with an ID.

Versions:

```
"loopback": "^2.18.0",
"loopback-boot": "^2.8.1",
"loopback-connector-mongodb": "^1.11.1",
"loopback-connector-mysql": "^2.1.1",
"loopback-datasource-juggler": "^2.30.0",
```

My update script:

```
Person.find(function(err, everyone) {
  if(err) {
    console.log(err);

  } else {
    var cnt = 0;

    _.forEach(everyone, function(person) {

      console.log(person); // id is present!

      var existingHash = person.hash;

      person.hash = getHash(person.email);

      person.save(function(err, personSaved) {
        if(err) {console.log(err);}

        cnt += 1;

        if(cnt === everyone.length) {
          mysqlDS.disconnect();
          process.exit(0);
        }

      });

    });

  }

});
```

My before save hook:

```
Person.observe('before save', function(ctx, next) {

  console.log('[before save] ctx.isNewInstance', ctx.isNewInstance);

  next();

});
```

My after save hook:

```
Person.observe('after save', function(ctx, next) {

  console.log('[after save] ctx.isNewInstance', ctx.isNewInstance);

  next();

});
```

Output when I run the script:

```
[before save] ctx.isNewInstance undefined
[after save] ctx.isNewInstance true
```

Shouldn't these both (explicitly) be false when calling `.save()` on a person entity with an id?

b-admike:
### Description
Fixes #106 

### Checklist

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -280,13 +280,13 @@ MySQL.prototype._modifyOrCreate = function(model, data, options, fields, cb) {
       data.id = info.insertId;
     }
     var meta = {};
-    if (info) {
-      // When using the INSERT ... ON DUPLICATE KEY UPDATE statement,
-      // the returned value is as follows:
-      // 1 for each successful INSERT.
-      // 2 for each successful UPDATE.
-      meta.isNewInstance = (info.affectedRows === 1);
-    }
+    // When using the INSERT ... ON DUPLICATE KEY UPDATE statement,
+    // the returned value is as follows:
+    // 1 for each successful INSERT.
+    // 2 for each successful UPDATE.
+    // 1 also for UPDATE with same values, so we cannot accurately
+    // report if we have a new instance.
+    meta.isNewInstance = undefined;
     cb(err, data, meta);
   });
 };
--- a/test/mysql.test.js
+++ b/test/mysql.test.js
@@ -4,10 +4,11 @@
 // License text available at https://opensource.org/licenses/MIT
 
 'use strict';
+var async = require('async');
 var should = require('./init.js');
 var sinon = require('sinon');
 
-var Post, PostWithStringId, PostWithUniqueTitle, PostWithNumId, db;
+var Post, PostWithStringId, PostWithUniqueTitle, PostWithNumId, Student, db;
 
 // Mock up mongodb ObjectID
 function ObjectID(id) {
@@ -57,18 +58,24 @@ describe('mysql', function() {
       content: {type: String},
     });
 
-    db.automigrate(['PostWithDefaultId', 'PostWithStringId', 'PostWithUniqueTitle', 'PostWithNumId'], function(err) {
-      should.not.exist(err);
-      done(err);
+    Student = db.define('Student', {
+      name: {type: String, length: 255},
+      age: {type: Number},
+    }, {
+      forceId: false,
     });
+
+    db.automigrate(
+      ['PostWithDefaultId', 'PostWithStringId',
+        'PostWithUniqueTitle', 'PostWithNumId', 'Student'],
+      function(err) {
+        should.not.exist(err);
+        done(err);
+      });
   });
 
-  beforeEach(function(done) {
-    Post.destroyAll(function() {
-      PostWithStringId.destroyAll(function() {
-        PostWithUniqueTitle.destroyAll(done);
-      });
-    });
+  beforeEach(function() {
+    return deleteAllModelInstances();
   });
 
   it('should allow array or object', function(done) {
@@ -232,6 +239,33 @@ describe('mysql', function() {
           should.not.exist(p._id);
           p.title.should.equal(post.title);
           p.content.should.equal(post.content);
+          done();
+        });
+      });
+    });
+
+    it('isNewInstance should be undefined for after save hook', function(done) {
+      var student = {name: 'Joe', age: 20};
+      var newStudent = {};
+      var isNewInstanceBefore = false;
+      var isNewInstanceAfter = false;
+      Student.create(student, function(err, createdStudent) {
+        if (err) return done(err);
+        newStudent.id = createdStudent.id;
+        newStudent.name = 'Hannah';
+        newStudent.age = 25;
+        Student.observe('before save', function(ctx, next) {
+          isNewInstanceBefore = ctx.isNewInstance;
+          next();
+        });
+        Student.observe('after save', function(ctx, next) {
+          isNewInstanceAfter = ctx.isNewInstance;
+          next();
+        });
+        Student.replaceOrCreate(newStudent, function(err, s) {
+          if (err) return done(err);
+          should.not.exist(isNewInstanceBefore);
+          should.not.exist(isNewInstanceAfter);
           done();
         });
       });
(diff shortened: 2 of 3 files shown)


b-admike:
@strongloop/sq-lb-apex @bajtos PTAL

test/mysql.test.js
```diff
@@ -235,6 +245,31 @@ describe('mysql', function() {
         });
       });
     });
+
+    it.only('isNewInstance should be undefined for after save hook', function(done) {
```
ssh24: Remove the `.only` :)

lib/mysql.js
```diff
@@ -280,13 +281,13 @@ MySQL.prototype._modifyOrCreate = function(model, data, options, fields, cb) {
...
     var meta = {};
-    if (info) {
         // When using the INSERT ... ON DUPLICATE KEY UPDATE statement,
         // the returned value is as follows:
         // 1 for each successful INSERT.
         // 2 for each successful UPDATE.
-      meta.isNewInstance = (info.affectedRows === 1);
-    }
+        // 1 also for UPDATE with same values, so we cannot accurately
+        // report if we have a new instance.
```
bajtos: The indentation seems to be off.

test/mysql.test.js
```diff
@@ -56,7 +56,15 @@ describe('mysql', function() {
...
-    db.automigrate(['PostWithDefaultId', 'PostWithStringId', 'PostWithUniqueTitle', 'PostWithNumId'], function(err) {
+    Student = db.define('Student', {
+      name: {type: String, length: 255},
+      age: {type: Number},
+    }, {
+      forceId: false,
+    });
+
+    db.automigrate(['PostWithDefaultId', 'PostWithStringId',
+    'PostWithUniqueTitle', 'PostWithNumId', 'Student'], function(err) {
```
bajtos: nitpick: http://loopback.io/doc/en/contrib/style-guide.html#one-argument-per-line

```js
db.automigrate(
 [...],
 function(err) {
 });
```

test/mysql.test.js
```diff
@@ -65,7 +73,9 @@ describe('mysql', function() {
   beforeEach(function(done) {
     Post.destroyAll(function() {
       PostWithStringId.destroyAll(function() {
-        PostWithUniqueTitle.destroyAll(done);
+        PostWithUniqueTitle.destroyAll(function() {
+          Student.destroyAll(done);
```
bajtos: Maybe it's time to use something like `async.each`?

```js
async.each(
  [Post, PostWithStringId/*...*/],
  function(Model, next) { Model.destroyAll(next); },
  done);
```

test/mysql.test.js
```diff
@@ -235,6 +245,31 @@ describe('mysql', function() {
...
+        newStudent.name = 'Hannah';
+        newStudent.age = 25;
+        Student.replaceOrCreate(newStudent, function(err, s) {
+          if (err) return done(err);
+          s.id.should.equal(newStudent.id);
+          s.name.should.equal(newStudent.name);
+          s.age.should.equal(newStudent.age);
+        });
+
+        Student.observe('before save', function(ctx, next) {
```
bajtos: This is relying on the fact that before/after save hooks are invoked asynchronously, not in the same tick as `Student.replace()` was called. Also if these hook as never invoked, then the test will still pass.

A better solution is to:
 - register the hooks before calling `replaceOrCreate`
 - in the hooks, save the observed context objects
 - in `replaceOrCreate` callback, assert that context objects were stored and have expected `isNewInstance` value.

test/mysql.test.js
```diff
@@ -878,7 +911,9 @@ describe('mysql', function() {
   after(function(done) {
     Post.destroyAll(function() {
       PostWithStringId.destroyAll(function() {
-        PostWithUniqueTitle.destroyAll(done);
+        PostWithUniqueTitle.destroyAll(function() {
+          Student.destroyAll(done);
```
bajtos: Ditto, use `async.each`? This function seems to be the same as the one above, can you extract it into a shared one please?

test/mysql.test.js
b-admike: I am not sure how to make sure they are called in the same tick as `Student.replaceOrCreate()` here. Should I call them synchronously in the following order (using callbacks or async.series)?
`Student.observe('before save')` -> `Student.observe('after save') -> `Student.replaceOrCreate`

bajtos: > Should I call them synchronously in the following order 

Yes, that's what I was proposing.

The point is that you need to call `Student.observe` before `Student.replaceOrCreate`. Because `observe()` is a synchronous method, there is no need to use callbacks/`async.series`.

```js
Student.observe('before save', ...);
Student.observe('after save', ...);
Student.replaceOrCreate(..., function(err) {
  // assert
  done();
});
```

test/mysql.test.js
```diff
@@ -21,6 +22,13 @@ ObjectID.prototype.toJSON = function() {
   return this.id1 + this.id2;
 };
 
+var deleteModels = function(models, done) {
+  async.each(
+      models,
+      function(Model, next) { Model.destroyAll(next); },
+        done);
```
bajtos: I think this line is indented too much, you should remove two spaces.

Also `deleteModel` should be defined as a regular function, the code should be placed towards the end of the `describe` block, after the tests. See http://loopback.io/doc/en/contrib/style-guide.html#layout-of-test-files

BTW since we can use ES6 and promises in `master`, the code can be simplified even more:

```js
// in your hook
beforeEach(function() {
  return deleteAllModelInstance([...]);
});

// function definition
function deleteAllModelInstances(models) {
  return Promise.all(models.map(m => m.destroyAll());
}
```

AFAICT, the list of `models` we are deleting is always the same, isn't it? In which case we can make `deleteAllModelInstance` parameter-less, which was what I had originally in mind.

```js
// at the top
beforeEach(deleteAllModelInstances);

// at the bottom
function deleteAllModelInstances() {
  const models = [...];
  return Promise.all(models.map(m => m.destroyAll());
}

// or perhaps
function deleteAllModelInstances() {
  return Promise.all([
    PostWithDefaultId.destroyAll(),
    // ...
    Student.destroyAll(),
  ]);
```

Thoughts?

Let's not waste too much time on this code cleanup - pick whatever makes most sense to you.

test/mysql.test.js
```diff
@@ -235,6 +249,36 @@ describe('mysql', function() {
...
+        });
+        Student.observe('after save', function(ctx, next) {
+          isNewInstanceAfter = ctx.isNewInstance;
+          next();
+        });
+        Student.replaceOrCreate(newStudent, function(err, s) {
+          if (err) return done(err);
+          s.id.should.equal(newStudent.id);
+          s.name.should.equal(newStudent.name);
+          s.age.should.equal(newStudent.age);
```
bajtos: I think these three asserts (L273-L275) can be removed - the test is checking how `ctx.isNewInstance` is reported. The fact that `replaceOrCreate` modifies the data correctly is already verified by other tests.

bajtos:
@b-admike IIUC, the tests are failing because the change adding `connectorCapabilities. replaceOrCreateReportsNewInstance` to juggler was not landed/released yet. I think it will be best to hold this pull request open until juggler is updated and the CI is passing.

b-admike:
@bajtos Thank you for your reviews. I have applied all your feedback, but I think this PR can go in the other direction as well. The `message` key of the OkPacket we get back from our driver can help us distinguish between a new instance and an old one if we turn off `CLIENT_FOUND_ROWS` flag. However, it is currently an empty string when we retrieve it via our connector as opposed to just using the driver, so I am looking into why that is.

b-admike:
@slnode test please

b-admike:
@slnode test please

PR #319, opened 2017-08-03
Use MySQL DEFAULT Clause/Constant

### LINKED ISSUE #318: Use MySQL DEFAULT Clause/Constant
# Description/Steps to reproduce
Adding the ability to use the built in MySQL `DEFAULT` Constraint.

```js
    ...
    "status": {
      "type": "string",
      "mysql": {
        "default":"pending"
      }
    },
    "last_modified": {
      "type": "date",
      "required": true,
      "mysql": {
        "default":"CURRENT_TIMESTAMP"
      }
    }
    ...
```

# Expected result
```bash
mysql> describe my_model_table;
+--------------------------+--------------+------+-----+-------------------+----------------+
| Field                    | Type         | Null | Key | Default           | Extra          |
+--------------------------+--------------+------+-----+-------------------+----------------+
| status                   | varchar(255) | YES  | MUL | pending           |                |
| last_modified            | datetime     | YES  |     | CURRENT_TIMESTAMP |                |
+--------------------------+--------------+------+-----+-------------------+----------------+
8 rows in set (0.00 sec)
```

ericdowell:
### Description
Adding mysql.default option and documentation on special case for date.

fixes #318 

#### Related issues

- strongloop/loopback-connector-mysql#318

### Checklist

- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,11 @@ For the date or timestamp types use `CURRENT_TIMESTAMP` or `now`:
   }
 }
 ```
+**NOTE**: The follow column types do **NOT** supported [MySQL Default Values](https://dev.mysql.com/doc/refman/5.7/en/data-type-defaults.html):
+- BLOB
+- TEXT
+- GEOMETRY
+- JSON
 
 ### Floating-point types
 
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ For the date or timestamp types use `CURRENT_TIMESTAMP` or `now`:
   }
 }
 ```
-**NOTE**: The follow column types do **NOT** supported [MySQL Default Values](https://dev.mysql.com/doc/refman/5.7/en/data-type-defaults.html):
+**NOTE**: The following column types do **NOT** supported [MySQL Default Values](https://dev.mysql.com/doc/refman/5.7/en/data-type-defaults.html):
 - BLOB
 - TEXT
 - GEOMETRY
--- a/README.md
+++ b/README.md
@@ -258,6 +258,26 @@ Use the `limit` option to alter the display width. Example:
 }
 ```
 
+### Default Clause/Constant
+Use the `default` property to have MySQL handle setting column `DEFAULT` value.
+```javascript
+"status": {
+  "type": "string",
+  "mysql": {
+    "default":"pending"
+  }
+}
+```
+For the date or timestamp types use `CURRENT_TIMESTAMP` or `now`:
+```javascript
+"last_modified": {
+  "type": "date",
+  "mysql": {
+    "default":"CURRENT_TIMESTAMP"
+  }
+}
+```
+
 ### Floating-point types
 
 For Float and Double data types, use the `precision` and `scale` options to specify custom precision. Default is (16,8). For example:
--- a/README.md
+++ b/README.md
@@ -264,7 +264,13 @@ Use the `default` property to have MySQL handle setting column `DEFAULT` value.
 "status": {
   "type": "string",
   "mysql": {
-    "default":"pending"
+    "default": "pending"
+  }
+},
+"number": {
+  "type": "number",
+  "mysql": {
+    "default": 256
   }
 }
 ```
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -701,13 +701,7 @@ function mixinMigration(MySQL, mysql) {
     var p = this.getModelDefinition(model).properties[prop];
     var line = this.columnDataType(model, prop) + ' ' +
       (this.isNullable(p) ? 'NULL' : 'NOT NULL');
-    if (typeof p.mysql !== 'undefined' && p.mysql.default && typeof p.mysql.default === 'string') {
-      line = line + ' DEFAULT ' +
-            (p.mysql.default.toUpperCase() === 'CURRENT_TIMESTAMP' || p.mysql.default.toLowerCase() === 'now' ?
-              'CURRENT_TIMESTAMP' : '"' + p.mysql.default + '"');
-      return line;
-    }
-    return line;
+    return columnDefault(p, line);
   };
 
   // override this function from base connector to allow mysql connector to
@@ -775,6 +769,28 @@ function mixinMigration(MySQL, mysql) {
     }
     return dt;
   };
+
+  function columnDefault(p, line) {
+    if (typeof p.type === 'undefined' || typeof p.type.name !== 'string') {
+      return line;
+    }
+    // Unsupported column types
+    if (['blob', 'json', 'text'].indexOf(p.type.name.toLowerCase()) !== -1) {
+      return line;
+    }
+    if (typeof p.mysql !== 'undefined' && p.mysql.default) {
+      var columnDefault = p.mysql.default;
+      if (typeof columnDefault === 'number') {
+        return line + ' DEFAULT ' + columnDefault;
+      }
+      if (typeof columnDefault === 'string') {
+        return line + ' DEFAULT ' +
+          (columnDefault.toUpperCase() === 'CURRENT_TIMESTAMP' || columnDefault.toLowerCase() === 'now' ?
+            'CURRENT_TIMESTAMP' : '"' + columnDefault + '"');
+      }
+    }
+    return line;
+  }
 
   function columnType(p, defaultType) {
     var dt = defaultType;
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -701,6 +701,12 @@ function mixinMigration(MySQL, mysql) {
     var p = this.getModelDefinition(model).properties[prop];
     var line = this.columnDataType(model, prop) + ' ' +
       (this.isNullable(p) ? 'NULL' : 'NOT NULL');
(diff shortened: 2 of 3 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

ssh24:
There is something similar in the loopback core level where you can specify a default value for a model property. Read [here](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html) and look for **defaultFn**.

ericdowell:
@ssh24 Thanks you for the link. I have read all the available ways to set a default value for a loopback model. 

Using `"defaultFn": "now"` to work around using `CURRENT_TIMESTAMP` on the database level does get the job done. But we should still have an interface to allow the responsibility to be placed on the database rather than in the software. This is also why I placed the value within the `mysql` key, this was done intentionally to show that it's being set via MySQL and not via loopback core:
```javascript
"status": {
  "type": "string",
  "mysql": {
    "default":"pending"
  }
}
```

ssh24:
I don't think this PR is needed. You can just specify the **default** value of the property on a property level. 

For example:
```json
  "properties": {
    "name": {
      "type": "string",
      "default": "Bob"
    },
    "age": {
      "type": "number"
    }
  },
```

Any other specific intention for this PR? From my understanding you are trying to achieve the same as I specified.

b-admike:
@ssh24 I think the value added from this PR is that it creates the table with a default value for the specified columns, so that if we leave the columns blank, MySQL will use the default value instead of doing an insert with the default value every time.

ericdowell:
@b-admike You're correct, also I've added unit test in https://github.com/strongloop/loopback-connector-mysql/pull/319/commits/de86e4782aa093301d00f5d3fd259c08e21b4b7d. Are they what you were looking for?

@ssh24 The intention of these changes are to allow the database do what the database does best.
As I put in the ticket (#318) this will allow for MySQL to take on setting the default value:
```shell
mysql> describe my_model_table;
+--------------------------+--------------+------+-----+-------------------+----------------+
| Field                    | Type         | Null | Key | Default           | Extra          |
+--------------------------+--------------+------+-----+-------------------+----------------+
| status                   | varchar(255) | YES  | MUL | pending           |                |
| last_modified            | datetime     | YES  |     | CURRENT_TIMESTAMP |                |
+--------------------------+--------------+------+-----+-------------------+----------------+
8 rows in set (0.00 sec)
```

ericdowell:
@ssh24 Another example of why you'd want to put the default value setting on the database level is for when connecting to the database with a non-loopback app where the auto-magic software default setting feature doesn't exist.

ssh24:
@ericdowell I see. Yes, after a bit of more research I figured out loopback does not do it the **correct** way for storing a default value for each connector. It just inserts the value for that field by default if it is undefined.

I am okay with this feature. Could we also get the **default** value back as part of discovering the schema? It would be the discovery side of this feature and I think it should support both migration and discovery.

Feel free to open a new PR if you want to do the discovery as a separate state.

ericdowell:
@ssh24 I can look into adding the default value as part of the the discovering schema at a later time in a separate PR. I'll open a ticket after this is merged as a reminder.

lib/migration.js
```diff
@@ -701,6 +701,12 @@ function mixinMigration(MySQL, mysql) {
     var p = this.getModelDefinition(model).properties[prop];
     var line = this.columnDataType(model, prop) + ' ' +
       (this.isNullable(p) ? 'NULL' : 'NOT NULL');
+    if (typeof p.mysql !== 'undefined' && p.mysql.default && typeof p.mysql.default === 'string') {
+      line = line + ' DEFAULT ' +
+            (p.mysql.default.toUpperCase() === 'CURRENT_TIMESTAMP' || p.mysql.default.toLowerCase() === 'now' ?
+              'CURRENT_TIMESTAMP' : '"' + p.mysql.default + '"');
```
ssh24: I don't think we should stringify the default value if it is not `CURRENT_TIMESTAMP`. There is a bunch of different other functions that can be specified as a default value. See [here](https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html)

ericdowell: @ssh24 Reading the `Default` bullet point on this documentation page: https://dev.mysql.com/doc/refman/5.7/en/create-table.html it states:

>Specifies a default value for a column. With one exception, the default value must be a constant; it cannot be a function or an expression. This means, for example, that you cannot set the default for a date column to be the value of a function such as NOW() or CURRENT_DATE. The exception is that you can specify CURRENT_TIMESTAMP as the default for a TIMESTAMP or DATETIME column. See [Section 11.3.5, “Automatic Initialization and Updating for TIMESTAMP and DATETIME”](https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html).

ericdowell: It also says these types aren't support:
>BLOB, TEXT, and JSON columns cannot be assigned a default value.

I can add logic to prevent these types from attempting to set a default.

ssh24: > It also says these types aren't supported:
`BLOB, TEXT, and JSON columns cannot be assigned a default value.`

> I can add logic to prevent these types from attempting to set a default.

Yes please.

Also my original concern was that we stringify all values. What if I have an integer column and want my default value to be 10? The current code will always insert it as `"10"` and throw error.

Can we also add properties of each data type with default values and add test cases for it?

ericdowell: I will add a `typeof` check for the value being a `number` and not stringify the value in this case.

lib/migration.js
```diff
@@ -770,6 +770,28 @@ function mixinMigration(MySQL, mysql) {
     return dt;
   };
 
+  function columnDefault(p, line) {
+    if (typeof p.type === 'undefined' || typeof p.type.name !== 'string') {
```
ssh24: Is this additional check required?
```
typeof p.type.name !== 'string
```

IMO, it will be causing issues with Number fields.

ssh24:
@slnode test please

lib/migration.js
ericdowell: @ssh24 Looking at the flow of the code I'm not sure I know what you mean by this. I've also written a [test for the number type field](https://github.com/ericdowell/loopback-connector-mysql/blob/4da3d375a757452dbad815f28db98018eb7b881b/test/migration.test.js#L389)

Here is the flow of the code that I can follow and why this needs to be checked:
A lot of things build up then [MySQL.prototype.buildColumnDefinition](https://github.com/ericdowell/loopback-connector-mysql/blob/4da3d375a757452dbad815f28db98018eb7b881b/lib/migration.js#L700) is called which then calls [this.columnDataType](https://github.com/ericdowell/loopback-connector-mysql/blob/4da3d375a757452dbad815f28db98018eb7b881b/lib/migration.js#L702) which calls [this.buildColumnType](https://github.com/ericdowell/loopback-connector-mysql/blob/4da3d375a757452dbad815f28db98018eb7b881b/lib/migration.js#L731) and within `buildColumnType` there's a reference to`p.type.name` in [this switch statement](https://github.com/ericdowell/loopback-connector-mysql/blob/4da3d375a757452dbad815f28db98018eb7b881b/lib/migration.js#L737-L769). 

From this flow and usage I'm lead to believe that a string value should be set as `p.type.name`. This isn't always true as without this check 9 other tests fail because `typeof p.type.name` is equal to `undefined`.

ssh24: Ah, I see. Discard my comment above.

README.md
```diff
@@ -258,6 +258,32 @@ Use the `limit` option to alter the display width. Example:
...
+```
+For the date or timestamp types use `CURRENT_TIMESTAMP` or `now`:
+```javascript
+"last_modified": {
+  "type": "date",
+  "mysql": {
+    "default":"CURRENT_TIMESTAMP"
+  }
+}
+```
```
b-admike: Can you document the unsupported types here? It'd also be good to throw a warning when users specify defaults for those types.

ericdowell: Added in this commit: https://github.com/strongloop/loopback-connector-mysql/pull/319/commits/97009acc18a75ee02243e4c9d159a8e5a9a35dd9

ericdowell: and fixed wording here: https://github.com/strongloop/loopback-connector-mysql/pull/319/commits/fa33dfa69e7cdd515c5fcf60ed7036464d15921f 😂

ssh24:
@slnode test please

ssh24:
@ericdowell PR is ready to land. Thank you again for the feature contribution.

I have created a story to track the discovery part. See [here](https://github.com/strongloop/loopback-connector-mysql/issues/320). Feel free to take leadership on adding this feature.

ericdowell:
@ssh24 Will these changes be tagged as a release soon?

ssh24:
@ericdowell I have added it to the release list. Should be out in the upcoming week.

PR #335, opened 2017-10-20
Lock versions of loopback-connector and mysql

rurivera6:
### Description
I started getting this error unless I locked down the versions of loopback-connector and mysql.  The new versions of loopback-connector 4.4.0 and mysql 2.15.0 were giving me an error along the lines of:
-----------------------------------------------------------------------------------
>> Generated Angular services file "client/app/core/shared/services/lbServices.js"
[SyntaxError: Cannot create data source "jess": /home/ubuntu/amp/node_modules/loopback-connector-mysql/node_modules/os-locale/index.js:2
const execa = require('execa');
^^^^^
Use of const in strict mode.]
-----------------------------------------------------------------------------------
When I compared to a different environment where my app works I noticed the difference in the version of the dependent modules.

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,8 @@
     "async": "^0.9.0",
     "debug": "^3.1.0",
     "lodash": "^4.17.4",
-    "loopback-connector": "^4.0.0",
-    "mysql": "^2.11.1",
+    "loopback-connector": "4.3.0",
+    "mysql": "2.14.1",
     "strong-globalize": "^3.1.0"
   },
   "devDependencies": {


slnode:

slnode:

slnode:

slnode:

dhmlau:
I'm wondering if the error will be fixed by updating the `os-locale` version than locking the `loopback-connector` and `mysql` versions.

rurivera6:
os-locale is included as part of strong-globalize and it looks like you made a change on Oct 17th (last week) to upgrade the version of it.
"strong-globalize": "^2.5.8" => "strong-globalize": "^3.1.0"
The version of os-locale that works is "os-locale@1.4.0" per my environment.  Are you thinking that the change to fix this is to lock down the version of os-local within strong-globalize@3.1.0 to depend on version 1.4.0 of os-locale?

dhmlau:
Which node version you're using?

rurivera6:
Sorry to get back so late on this.  The node version I am using is v0.12.18.

rurivera6:
So ignore my previous comment regarding your updating of strong-globalize because I'm using version loopback-connector-mysql@5.1.0 and that doesn't have your update.  But back to your suggestion with regards to the os-locale, I did make that change within strong-globalize to lock down the os-locale to 1.4.1 and that worked and fixed that issue.  I did run into a similar issue with the same type of error message and had to also lock down the version of g11n-pipeline to 1.5.1.  It's a pain dealing with all these issues but at least I have a way of fixing them for now.  However, every time I roll out a new package of my software to any environment I will have to manually fix these issues.

rurivera6:
I guess we can close this PR and open up a new one in strong-globalize to lock down the os-locale to 1.4.0 and g11n-pipeline to 1.5.1.

Issue #345, opened 2018-01-19
Error in hitting queries with LB_LAZYCONNECT_DATASOURCES option as true


akki-ng:
Loopback provides an option of lazy connect to datasource by setiing ```LB_LAZYCONNECT_DATASOURCES```.

While the same works on ```loopback-connector-mongo```, but ```loopback-connector-mysql``` gives the error.
```
0|onboardi | Unhandled error for request GET /api/Orders/paginate?access_token=RZisJvfXT4Gfm0SCtUodDpEIvnJaZb5U17S6RKzQ5lulSVVqdUIZ0WcB4mKLibYU: TypeError: Cannot read property 'escapeId' of undefined
0|onboardi |     at MySQL.escapeName (/home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-connector-mysql/lib/mysql.js:477:22)
0|onboardi |     at MySQL.SQLConnector.columnEscaped (/home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-connector/lib/sql.js:457:15)
0|onboardi |     at /home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-connector/lib/sql.js:1355:17
0|onboardi |     at Array.map (<anonymous>)
0|onboardi |     at MySQL.SQLConnector.buildColumnNames (/home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-connector/lib/sql.js:1354:20)
0|onboardi |     at MySQL.SQLConnector.buildSelect (/home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-connector/lib/sql.js:1376:10)
0|onboardi |     at MySQL.find [as all] (/home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-connector/lib/sql.js:1439:19)
0|onboardi |     at invokeConnectorMethod (/home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-datasource-juggler/lib/dao.js:169:21)
0|onboardi |     at /home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-datasource-juggler/lib/dao.js:2129:7
0|onboardi |     at doNotify (/home/akshay/NotInQ/git/onboarding-platform/node_modules/loopback-datasource-juggler/lib/observer.js:155:49)
[...]

PR #346, opened 2018-01-19
lazyConnect fix

akki-ng:
It will fix this issue. #345 
I'm not sure if this is the only place where we need to correct it, so opening a PR for discussion.

### DIFF
--- a/lib/mysql.js
+++ b/lib/mysql.js
@@ -469,6 +469,12 @@ MySQL.prototype.fromColumnValue = function(prop, val) {
  * @returns {string} The escaped database identifier
  */
 MySQL.prototype.escapeName = function(name) {
+  if (this.settings.lazyConnect && !this.client) {
+    var self = this;
+    var options = generateOptions(this.settings);
+    var s = self.settings || {};
+    this.client = mysql.createPool(options);
+  }
   return this.client.escapeId(name);
 };
 


slnode:

akki-ng:
@jannyHou Sure, I'll give you a scenario where ```lazy_connect``` is helpful. In my ```datasources.js```, I've defined ```memory, transient,  mongodb, mysql, email, mgage, google maps, storage-component1, storage-component2, onesignal```

As you can see I've lots of datasources defined. And I'm able to run things perfectly using `node .`
As soon as I go for cluster mode or with any standard ```process-manager```, MYSQL throws an error ```"Connection closed by server"```

After delving deep inside ```loopback-boot```, I realised that datasources are being initialised instantly in a for-loop without using ```async-map```. So as we go with cluster all cluster instances are jamming mysql. And another problem is mysql protocol

1. Handshake
2. Send Authentication Data <within some time window after handshake>
2. Connected <Use connection within some time window else idle/killed>

In this scenario, node js typically being single threaded is not giving cpu juice to ```mysqljs/mysql``` in time, which throws an error based on mysql protocol ```connection closed by server```

I will definitely not want to use insanely huge values for ```connectTimout, accquireTimeout```.

Another problem is, if I want to boot my app faster with these many datasources, it takes significant amount of time, So using ```lazy_connect``` I'm delaying initialization of each datasource until it is required, for which ```loopback-connector-mysql``` threw an error. This PR is to initialise ```loopback-connector-mysql```when it is required, making application boot up faster.

If you find it as a good reason let me know I'll clean up the code. Current PR needs to be cleaned up, was waiting to know if this is the only piece which needs to be patched.

akki-ng:
@jannyHou Any update on this?

dhmlau:
@slnode test please

dhmlau:
@akki-ng , thanks for pinging. 

The error is:
```
Error: Cannot find module 'loopback-datasource-juggler/test/common.batch.js'
```
@rmg , we have a similar issue with the mongodb connector.  You've helped us resolved.  Could you please help?  Thanks.
> the problem is the `loopback-datasource-juggler` package being installed doesn't actually have those files
> the version published to ci.strongloop.com

rmg:
@slnode test please

dhmlau:
@slnode test please
give it one more try. CI on node 8 passes now, but not node 6.

jannyHou:
@akki-ng Sorry for late reply I was away on another project last month and just back for loopback, if you want to delay the connection until a query sends to the datasource, you can try set `lazyConnect` to true in datasource settings, and then in your bootscript, reset them to false. Because the bootscript is executed after datasource booting phase.

Or the mysql error happens after you do something like above to skip the connection when booting datasource and re-enabled it afterwards? 

You can fork https://github.com/strongloop/loopback-sandbox to give us an app reproduce the error. That would get us on the same page faster then describing it in words. Thanks!

akki-ng:
@jannyHou I did set ```lazyConnect``` to true in datasources setting. I was not planning to hack the system by using bootscripts. Even if using bootscripts in the mentioned fashion will not solve the issue, as the ```client``` would not have been initialised. We must handle below conditions in our code base.
1. Lets not assume ```client``` would have been initialised
2. If ```client``` was not initialised and ```lazyConnect``` is set, then manually create a new ```client```
3. if ```client``` was not initialised and ```lazyConnect``` was false , then we must throw an error regarding ```client initialisation failure```.
Let me know your thoughts. Thanks.

dhmlau:
@jannyHou , could you please follow up? thanks!

akki-ng:
@dhmlau @jannyHou Any update on this?

Issue #359, opened 2018-09-10
Security issues with outdated dependency


tgijasonyip:
# Description/Steps to reproduce

There is a security vulnerability in "strong-globalize": "^3.1.0" which lead to the dependency of package - deep-extend@0.4.2. The version of the deep extend package has [Prototype Pollution](https://nodesecurity.io/advisories/612) issue.

![image](https://user-images.githubusercontent.com/38749833/45283125-a0b2a800-b50f-11e8-945e-b7b4fabbf409.png)

# Link to reproduction sandbox

Not apply for this issue.

# Expected result

No security vulnerabilities after running nsp check

# Additional information

There was a similar solved issue (Click [here](https://github.com/strongloop/loopback-connector-mongodb/issues/433)) reported in the loopback-connector-mongodb repository.

dhmlau:
We should be using `strong-globalize@4.1.1` instead of 3.x.  Let me take a look.

dhmlau:
Release in 5.3.1.

PR #360, opened 2018-09-11
update strong-globalize and mocha version

dhmlau:
### Description
Update strong-globalize and mocha version
Fixes https://github.com/strongloop/loopback-connector-mysql/issues/359

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,14 +18,14 @@
     "lodash": "^4.17.4",
     "loopback-connector": "^4.0.0",
     "mysql": "^2.11.1",
-    "strong-globalize": "^3.1.0"
+    "strong-globalize": "^4.1.1"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


dhmlau:
@virkt25 , per our discussion, after I updated mocha to 5.x, a few tests failed and I needed some more time to look into it.  Since `mocha` is one of the devDependencies instead of dependencies, you mentioned that it is less severe in the sense that mocha won't get installed in prod environment.  

As a conclusion, I'd like to only update the `strong-globalize` version in this PR and work on the mocha update (together with fixed test cases) in a separate PR.  If you agree, I'll modify the commit message and the title of this PR, and get it merged first.

virkt25:
@dhmlau That's a good idea. Lets land this and address `mocha` upgrade in a follow up PR.

Issue #366, opened 2018-11-01
Foreign Keys definition issue


fnatdev:
Hi,
I'm using loopback with MySQL and I've noted that's impossible to add the constraint "ON DELETE" and "ON UPDATE" to your foreign keys.
I've looked in code and I've found this function

`MySQL.prototype.buildForeignKeyDefinition = function(model, keyName) {
    var definition = this.getModelDefinition(model);

    var fk = definition.settings.foreignKeys[keyName];
    if (fk) {
      // get the definition of the referenced object
      var fkEntityName = (typeof fk.entity === 'object') ? fk.entity.name : fk.entity;

      // verify that the other model in the same DB
      if (this._models[fkEntityName]) {
        return ' CONSTRAINT ' + this.client.escapeId(fk.name) +
          ' FOREIGN KEY (`' + expectedColNameForModel(fk.foreignKey, definition) + '`)' +
          ' REFERENCES ' + this.tableEscaped(fkEntityName) +
          '(' + this.client.escapeId(fk.entityKey) + ')';
      }
    }
    return '';
  };`

As you can see the constraints is absent in SQL code generation. This is annoying if you use the autoupdate.
For the moment I removed the foreign keys from model definitions and added them manually manually using alter table statements.

Can you add this missing feature please!
Thanks!

Francesco

PR #363, opened 2018-09-17
#362 fix bug Foreign key constraint is incorrectly formed

myvo:
The issue #362 is fixed.
After checking in MySQL by command `SHOW ENGINE INNODB STATUS;`, I got the reason.
<pre>
LATEST FOREIGN KEY ERROR
------------------------
2018-09-17 10:40:11 0x7f2eb41b9700 Error in foreign key constraint of table `dbname`.`Book`:
Alter  table `lbmysql`.`Book` with foreign key constraint failed. Parse error in ' FOREIGN KEY (`author_id`) REFERENCES `Author`(`aId`)' near '`aId`)'.
</pre>

We forgot get the MySQL column name before building the Foreign key.

PR #370, opened 2018-12-04
Feature add foreign key on update on delete trigger config

HugoPoi:
### Description

Add possibility to configure the foreign key constraint with 2 optional triggers.
* add `onUpdate` foreign key settings
* add `onDelete` foreign key settings
* fix a issue, foreign key index try to be deleted on autoupdate 
* fix a issue, index being recreate each migration when declared directly on property

#### Related issues

* #366 
* I touch code in the same section than the PR #363

### Checklist

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -364,19 +364,20 @@ function mixinMigration(MySQL, mysql) {
       if (!found) {
         var colName = expectedColNameForModel(propName, m);
         var pName = self.client.escapeId(colName);
+        var indexName = self.client.escapeId(propName);
         var type = '';
         var kind = '';
         if (i.type) {
           type = 'USING ' + i.type;
         }
         if (kind && type) {
-          sql.push('ADD ' + kind + ' INDEX ' + pName +
+          sql.push('ADD ' + kind + ' INDEX ' + indexName +
             ' (' + pName + ') ' + type);
         } else {
           if (typeof i === 'object' && i.unique && i.unique === true) {
             kind = 'UNIQUE';
           }
-          sql.push('ADD ' + kind + ' INDEX ' + pName + ' ' + type +
+          sql.push('ADD ' + kind + ' INDEX ' + indexName + ' ' + type +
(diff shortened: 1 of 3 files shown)


slnode:

jannyHou:
@slnode test please

HugoPoi:
I'm not sure but some test in datasource juggler in windows node 6.x failed, is it related with my changes ?

derdeka:
@dhmlau @virkt25 @jannyHou @b-admike
Is this project dead / end of live as no one is reviewing the PR?
This PR solves several bugs in loopback-connector-mysql which prevent me to use loopback at all.

lib/migration.js
```diff
@@ -37,6 +37,35 @@ function mixinMigration(MySQL, mysql) {
...
+          if (match) {
+            actualFks.forEach(function(fk) {
+              if (fk.fkName === match[1]) {
+                fk.onDelete = match[5] ? match[5].toLowerCase() : 'restrict';
+                fk.onUpdate = match[6] ? match[6].toLowerCase() : 'restrict';
+              }
+            });
+            rawConstraints.push(match);
+          }
+        } while (match != null);
```
jannyHou: the `while` loop is a little bit risky to me, what if the statement doesn't contain the constraint? Maybe change the logic to stop when reaches the last line of the statement.

jannyHou:
@slnode test please

lib/migration.js
HugoPoi: the `if` statement line 54 in the `while` protect for no match in the `CREATE TABLE`, I think the tricky part is when someone rework my code later, because the iteration over the Regex is not obvious. Maybe add some comment before the line 52 to explain the iteration on the multiple match with RegExp object internal pointer.
Or is there a other way to iterate with complex RegExp ?

dhmlau:
@jannyHou, the CI is all green.  Do you think this PR is good to merge? Thanks.

jannyHou:
@slnode test please

PR #374, opened 2019-03-01
update lodash

jannyHou:
update lodash version


dhmlau:
I think we need to update the `.travis.yml` to get rid of the CI failure. It's using very old node.js versions: https://github.com/strongloop/loopback-connector-mysql/blob/master/.travis.yml

PR #376, opened 2019-04-18
chore: update travis and CODEOWNERS

dhmlau:
### Description
- Update `.travis.yml` to remove the node.js versions that are already EOL
- Update `CODEOWNERS` to reflect the current status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - 0.6
-  - 0.8
-  - 0.10
+  - "6"
+  - "8"
+  - "10"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - 0.6
-  - 0.8
-  - 0.10
```
b-admike: good catch! I'm surprised we haven't upgraded them until now.

dhmlau:
@nabdelgadir , Travis was failing before this PR, and the error is because of the database not found.

PR #379, opened 2019-05-05
 Fix AUTO_INCREMENT for complex PK

loryk:
### Description

In case we are use complex primary keys, one or many of them can be auto incremented fields. So I think need to add AUTO_INCREMENT without check PRIMIRY is simple or complex.

PR #380, opened 2019-05-05
Default value improvments

loryk:
### Description

MySQL server doesn't support boolean, but has bit, tinyint etc. 
Additional we can use empty default value, for example '' for strings, or 0 for int.  
The older code checks type of default value by type any way, so I think we can improve this case for:
 - boolean types
 - empty or 0 value
 - add support 'default' from generic properties (Like Postgres for example)

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -821,16 +821,25 @@ function mixinMigration(MySQL, mysql) {
     if (['blob', 'json', 'text'].indexOf(p.type.name.toLowerCase()) !== -1) {
       return line;
     }
-    if (typeof p.mysql !== 'undefined' && p.mysql.default) {
-      var columnDefault = p.mysql.default;
-      if (typeof columnDefault === 'number') {
-        return line + ' DEFAULT ' + columnDefault;
-      }
-      if (typeof columnDefault === 'string') {
-        return line + ' DEFAULT ' +
-          (columnDefault.toUpperCase() === 'CURRENT_TIMESTAMP' || columnDefault.toLowerCase() === 'now' ?
-            'CURRENT_TIMESTAMP' : '"' + columnDefault + '"');
-      }
+    var columnDefault = p.default;
+
+    if (typeof p.mysql !== 'undefined' && typeof p.mysql.default !== 'undefined')
+      columnDefault = p.mysql.default;
+
+    if (typeof columnDefault === 'boolean') 
+      return line + ' DEFAULT ' + Number(columnDefault);
+
+    if (typeof columnDefault === 'number') 
+      return line + ' DEFAULT ' + columnDefault;
+    
+    if (typeof columnDefault === 'string') {
+      if (columnDefault.toLowerCase() === '$now')
+        return line;
+        
+      return line + ' DEFAULT ' +
+        (columnDefault.toUpperCase() === 'CURRENT_TIMESTAMP' ||
+          columnDefault.toLowerCase() === 'now' ?
+          'CURRENT_TIMESTAMP' : '"' + columnDefault + '"');
     }
     return line;
   }


slnode:

dhmlau:
@loryk, thanks for another PR!  It seems like some of the changes in PR #379 are in here.  I believe you meant to have those separated? 

Also, could you please fix the commit linter error:
```
**
**  Linting commit logs
**
**  6 problems found:
**    88c34ea - Support 'default' value from general, and boolean t: First line should be 50 characters or less (saw 54)
**    88c34ea - Support 'default' value from general, and boolean t: line 4 longer than 72 characters.
**    88c34ea - Support 'default' value from general, and boolean t: line 5 longer than 72 characters.
**    88c34ea - Support 'default' value from general, and boolean t: line 8 longer than 72 characters.
**    631a93a - Merge pull request #1 from loryk/AUTO_INCREMENT_for: First line should be 50 characters or less (saw 63)
**    b5190c7 - Fix AUTO_INCREMENT for complex primary keys: line 3 longer than 72 characters.
**
```

Please see our commit message guideline for reference: https://loopback.io/doc/en/contrib/git-commit-messages.html

loryk:
Yes, I make first commit, and believe that is will be last, but after some time I found second issue)))
 I have some hope that first patch will be merged soon) 
So, it is not an issue. If patch not will be merged, I can split this fix from another)

loryk:
Done. (split commits, and fix commit message)

loryk:
For boolean value I see fails from juggler tests. 
Used both version 3.x and 4.x, so I should fix juggler project? 
Need to path 3.x and 4.x? 
How I should link this issue for juggler?

loryk:
Seems like need to add false in expected:
- strongloop/loopback-datasource-juggler#1768
But currently I can't check that this right.
Can any one helps me with fix tests?

loryk:
Any news?

bajtos:
Thank you for the pull request, @loryk. IIUC, you are changing the default value used by the database from `NULL` to `false`.

Now if you look at juggler tests that are failing with your changes in place (see https://github.com/strongloop/loopback-datasource-juggler/pull/1768), they are defining the boolean property as optional, i.e. allowing "no value":

```js
Player.defineProperty('active', {
  type: Boolean,
  default: false,
});
```

As I am interpreting such model definition, it says that `active` property/column should allow three possible values: `NULL`, `false` and `true`.

If we wanted to say that the property must be always set (i.e. can hold only `false` or `true`, but never `NULL`), then we would define it as follows:

```js
Player.defineProperty('active', {
  type: Boolean,
  required: true,
  default: false,
});
```

I am concerned that by landing your changes, we may introduce behavior that will confuse our users and possible break their existing applications (but this can be handled by releasing a semver-major version).

To move this pull request forward, I'd like you to do a bit of research and answer the following questions:
- How are other types handled my the MySQL connector? For example, when we define property as `{type: Number, default: 0}` and run autoupdate, will the database allow `NULL` to be stored in this column or not? How about strings?
- How are other databases (e.g. PostgreSQL and MongoDB) handling optional properties with a default value configured? Please check booleans, numbers and strings.

> Used both version 3.x and 4.x, so I should fix juggler project?
> Need to path 3.x and 4.x?
> How I should link this issue for juggler?

You will need to fix the issue in both 3.x and 4.x version of juggler. To verify the fix locally, use `npm link` to install your local version of juggler into your local mysql connector.

```shell
$ cd loopback-connector-mysql
$ npm link ../loopback-datasource-juggler
```

bajtos:
@loryk Also please add a test to verify your changes and prevent regressions in the future, see [test/mysql.autoupdate.test.js](https://github.com/strongloop/loopback-connector-mysql/blob/master/test/mysql.autoupdate.test.js)

loryk:
As I understand loopback approach, for not NULL values, need to use 'nullable' option and it is works fine. Additional check with 'default: false' only constrict  connector possibility. Before suggest this patch, I found different behaviour with PostgreSQL and MySQL connectors, and trying to explore this difference.

About your questions, 
I will try to double check 'autoupdate' when found little more time, but as for me 'nullable' and 'default' should not to be tie. 
About PostgreSQL and MongoDB: PostgreSQL works same as this patch (or very closely). For example I have project that use same models with both DB and autoupdate with this change provide equals DB structure. But about Mongo, I doesn't works with this type of DB, and I doesn't have enough  expertise to found this answer.
Unfortunately, right now I don't have necessary time to fix tests, but try look into as soon as possible.

Thank you for your attention.

bajtos:
Thank you @loryk for the answers. 

Here is the thing that I don't understand: if the PostgreSQL connector is working very similarly to what you are proposing here, then how comes that you need to change the juggler tests in https://github.com/strongloop/loopback-datasource-juggler/pull/1768 to make them pass with your changes in place?

Issue #367, opened 2018-11-03
Pool is closed if datasource.disconnect is called at any point before getConnection


tmcgee123:
# Description/Steps to reproduce

1. define a boot script like such that alls the dataSource to disconnect
```
// server/boot/create-lb-tables.js
module.exports = function(server, cb){
    var ds = server.dataSources.ping;
    var lbTables = ['AccessToken', 'ACL', 'RoleMapping', 'Role'];

    ds.autoupdate(lbTables, function(er){
        if (er) throw er;
        console.log('Loopback tables [' + lbTables + '] created in ', ds.adapter.name);
        ds.disconnect();
        cb();
    });
};
```
2. Run application via `node server.js`
3. make login request or run a script that does something with the same datasource
4. see that error is produced
```
2018-11-03T22:17:32.923Z loopback:datasource Datasource localhost: connected=false connecting=false
2018-11-03T22:17:32.923Z loopback:datasource Connecting datasource localhost to connector mysql
2018-11-03T22:17:32.930Z loopback:datasource Datasource localhost is now connected - executing method find
2018-11-03T22:17:32.932Z loopback:datasource Datasource localhost: connected=true connecting=false
2018-11-03T22:17:32.939Z loopback:juggler:utils Sanitizing query object: {"where":{"model":{"inq":["User","*"]},"property":{"inq":["login","*"]},"accessType":{"inq":["EXECUTE","*"]}}}
2018-11-03T22:17:32.947Z loopback:connector:mysql SQL: SELECT `model`,`property`,`accessType`,`permission`,`principalType`,`principalId`,`id` FROM `ACL` WHERE `model` IN (?,?) AND `property` IN (?,?) AND `accessType` IN (?,?) ORDER BY `id`, params: ["User","*","login","*","EXECUTE","*"]
2018-11-03T22:17:32.951Z loopback:connector:sql Error: {} "SELECT `model`,`property`,`accessType`,`permission`,`principalType`,`principalId`,`id` FROM `ACL` WHERE `model` IN (?,?) AND `property` IN (?,?) AND `accessType` IN (?,?) ORDER BY `id`" ["User","*","login","*","EXECUTE","*"]
Error: Pool is closed.
    at /var/www/ping-api/node_modules/loopback-connector-mysql/node_modules/mysql/lib/Pool.js:26:17
    at _combinedTickCallback (internal/process/next_tick.js:132:7)
[...]

PR #385, opened 2019-05-09
Fixes strongloop/loopback-connector-mysql#367

TwistedLogic:
### Description

Calling the `disconnect()` method and then the `connect()` method causes the "Pool is closed" error.

#### Related issues

- connect to strongloop/loopback-connector-mysql#367


slnode:

dhmlau:
@slnode test please

TwistedLogic:
@b-admike sure, I'll provide tests soon.

The current `ds.disconnect()` behavior is wrong regardless of placing it in a boot script. In my use case, I'm using loopback inside an AWS lambda, closing all the connections to the database at the end of each lambda function execution.

This fix solves the related issue too, I believe no disconnection should be made in that boot script though.

b-admike:
@TwistedLogic Thanks for adding tests! For some reason, the tests have not been run on `cis-jenkins`. @slnode test please.

b-admike:
> @b-admike sure, I'll provide tests soon.
> 
> The current `ds.disconnect()` behavior is wrong regardless of placing it in a boot script. In my use case, I'm using loopback inside an AWS lambda, closing all the connections to the database at the end of each lambda function execution.
> 
> This fix solves the related issue too, I believe no disconnection should be made in that boot script though.

Yeah I agree it should allow a re-connect in any case. Taking a look at your test locally (it's failing for me FYI):

```
  1060 passing (43s)
  105 pending
  4 failing

  1) connections should disconnect then connect and ORM should work:
     TypeError: Cannot read property 'escapeId' of null
      at MySQL.escapeName (lib/mysql.js:472:22)
      at MySQL.SQLConnector.tableEscaped (node_modules/loopback-connector/lib/sql.js:452:15)
      at MySQL.SQLConnector.dropTable (node_modules/loopback-connector/lib/sql.js:1544:47)
      at node_modules/loopback-connector/lib/sql.js:1592:10
      at node_modules/async/dist/async.js:3110:16
      at eachOfArrayLike (node_modules/async/dist/async.js:1069:9)
      at eachOf (node_modules/async/dist/async.js:1117:5)
      at Object.eachLimit (node_modules/async/dist/async.js:3172:5)
      at MySQL.SQLConnector.automigrate (node_modules/loopback-connector/lib/sql.js:1591:9)
      at DataSource.automigrate (node_modules/loopback-datasource-juggler/lib/datasource.js:1050:18)
      at Context.<anonymous> (test/connection.test.js:70:8)
```

test/connection.test.js
```diff
@@ -59,6 +59,38 @@ describe('connections', function() {
     });
   });
 
+  it('should disconnect then connect and ORM should work', function(done) {
```
b-admike: Let's rework this test based on Promises and be explicit in checking `ds.connected` value. I also used a new datasource instance to not pollute `db` used by other tests:

```js
  it('should disconnect then connect and ORM should work', function() {
    var ds = new DataSource(mysqlConnector, config);
    var Student = ds.define('Student', {
      name: {type: String, length: 255},
      age: {type: Number},
    }, {
      forceId: false,
    });

    return ds.connect()
      .then(function(err) {
        should.not.exist(err);
        return ds.automigrate(['Student']);
      })
      .then(function(err) {
        should.not.exist(err);
        should(ds.connected).be.True();
        return ds.disconnect();
      })
      .then(function(err) {
        should.not.exist(err);
        should(ds.connected).be.False();
        return ds.connect();
      })
      .then(function(err) {
        should.not.exist(err);
        should(ds.connected).be.True();
        return Student.count();
      })
      .then(function(count) {
        should(count).be.a.Number();
        return ds.disconnect();
      });
  });
```

TwistedLogic: Looks great, thanks. I didn't use promises to be compliant to the rest of the code.

b-admike: Yea that's a good point, but mocha supports both flavours and this is more elegant way of writing it. Also, before these changes, the linter was complaining of the `async function` callback you had for `automigrate` call (you can see this by running `npm run lint`).

TwistedLogic: I know, thanks for pointing me out. I always use async/await and promises in my code, also for testing using mocha. Sometimes, contributing to a project is a matter of adapting own personal style and conventions to those of the project. Anyway, your changes are great, thanks.

b-admike: I went into more details because I thought you were not convinced haha. I didn't know you were expecting me to make the changes. I was expecting you to apply them 😄. I'll be happy to make the changes but given that you have them in your `master` branch, I might not be able to do so. I'll give it a try and report back.

b-admike:
@TwistedLogic I've made the changes to your test, squashed the commits, and changed the commit message. Feel free to change it, but LGTM.

TwistedLogic:
@b-admike looks perfect, thanks

b-admike:
@slnode test please

dhmlau:
@TwistedLogic, thanks for the PR.  Could you please fix the commit linter error? 
```
**
**  Linting commit logs
**
**  1 problems found:
**    6856ebe - fix: allow DataSource to reconnect after disconnect: First line should be 50 characters or less (saw 54)
**
```

As a reference, here is our commit message guideline: https://loopback.io/doc/en/contrib/git-commit-messages.html.

b-admike:
@dhmlau Sorry that was me. Fixed.

b-admike:
@slnode test please

dhmlau:
@slnode test please

dhmlau:
@TwistedLogic , thanks for your contribution. Your PR has landed! 🎉

TwistedLogic:
@dhmlau Happy to contribute ;)

pharindoko:
@TwistedLogic
Can you do me a favor and explain with a codesnippet how and when you connect and disconnect the datasource when running the code in aws lambda.  With the freeze / thaw process of lambda this is pretty tricky :-/

TwistedLogic:
@pharindoko 

```javascript
// Booting loopback outside the function makes the lambda
// function execute faster when containers get re-used
const app = require('./server/server')

module.exports.func = async () => {

  // Your function code here
  // When you call the ORM methods the connection pool should use fresh connections.

  // **We need to close all the connections to the database before the lambda function returns.**
  // Lambda containers stay idle for 15 minutes when not used.
  // Max 1000 containers can be allocated for each lambda function which generates
  // too many open and useless connections to the database.
  // Moreover, when the container is destroyed, the connection is not properly closed and
  // MySQL increments max_connect_errors. Lambda IPs can be blocked by MySQL for this reason.
  await new Promise((resolve, reject) => {
    app.dataSources.mysql.disconnect(err => {
      if (err) return reject(err)
      resolve()
    })
  })

  return {
    statusCode: 204
  }
}
```

bajtos:
I believe that `disconnect` already supports promise, thus the code above can be simplified as follows:

```ts
module.exports.func = async () => {

  // Your function code here
  // When you call the ORM methods the connection pool should use fresh connections.

  // We need to close all the connections to the database before
  // the lambda function returns.
  // Lambda containers stay idle for 15 minutes when not used.
  // Max 1000 containers can be allocated for each lambda function which generates
  // too many open and useless connections to the database.
  // Moreover, if the container is destroyed and the connection is still alive,
  // the connection is not properly closed and MySQL increments max_connect_errors.
  // Lambda IPs can be blocked by MySQL for this reason.
  await app.dataSources.mysql.disconnect();

  return {
    statusCode: 204
  }
}
```

See https://github.com/strongloop/loopback-datasource-juggler/blob/5e3e1198d6d36f588f9c15237c6056b09c93be3b/lib/datasource.js#L2131-L2146 (in 3.x branch):

```js
DataSource.prototype.disconnect = function disconnect(cb) {
  cb = cb || utils.createPromiseCallback();
  // ...
  return cb.promise;
};
```

Issue #378, opened 2019-04-22
Travis CI fails


dhmlau:
Travis CI is failing with the following error.  There might be some configuration we need to do. 

```
> loopback-connector-mysql@5.3.1 pretest /home/travis/build/strongloop/loopback-connector-mysql
> node pretest.js
seeding DB with example db...
Warning: Using a password on the command line interface can be insecure.
ERROR 1045 (28000): Access denied for user 'test'@'localhost' (using password: YES)
done seeding DB
npm ERR! Test failed.  See above for more details.
The command "npm test" exited with 1.
```

dhmlau:
Closing as done 🎉 .

PR #387, opened 2019-05-21
change config of env variables of travis

agnes512:
#### Related issues

- connect to https://github.com/strongloop/loopback-connector-mysql/issues/378

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,19 @@ node_js:
   - "6"
   - "8"
   - "10"
-before_script:
-  - "mysql -e 'create database myapp_test;'"
+# Xenial build environment comes with MySQL 5.7
+dist: xenial
+services:
+  - mysql
+env:
+  global:
+    - MYSQL_HOST=localhost
+    - MYSQL_PORT=3306
+    - MYSQL_USER=test
+    - MYSQL_PASSWORD=test
+before_install:
+  - mysql -u root -e "CREATE USER 'test'@'localhost' IDENTIFIED BY 'test';"
+  - mysql -e "use mysql; update user set authentication_string=PASSWORD('test') where User='test'; update user set plugin='mysql_native_password';FLUSH PRIVILEGES;"
+  - mysql -e "GRANT ALL PRIVILEGES ON *.* TO 'test'@'localhost' WITH GRANT OPTION;FLUSH PRIVILEGES;"
+  - mysql -e "GRANT SUPER ON *.* TO 'test'@'localhost' IDENTIFIED BY 'test';FLUSH PRIVILEGES;"
+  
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,5 +3,20 @@ node_js:
...
+    - TEST_MYSQL_USER=test
+    - TEST_MYSQL_PASSWORD=test
+    - MYSQL_HOST=localhost
+    - MYSQL_PORT=3306
+    - MYSQL_USER=test
+    - MYSQL_PASSWORD=test
 before_script:
-  - "mysql -e 'create database myapp_test;'"
+  - mysql -u root -e "CREATE USER 'test'@'localhost' IDENTIFIED BY 'test';"
+  - mysql -e "use mysql; update user set authentication_string=PASSWORD('test') where User='test'; update user set plugin='mysql_native_password';FLUSH PRIVILEGES;"
```
dhmlau: I wonder where the database `mysql` is created.  
From this line https://github.com/strongloop/loopback-connector-mysql/blob/master/test/init.js#L17, it seems like we are using `myapp_test` by default?

jannyHou: looking at https://docs.travis-ci.com/user/database-setup/#mysql, looks like `mysql` is a default schema for user/system management.

`myapp_test` is the schema created for tests.
cc @dhmlau just my understanding :-p

.travis.yml
```diff
@@ -3,5 +3,21 @@ node_js:
   - "6"
   - "8"
   - "10"
+services:
+  - mysql
+env:
+  global:
+    # - TEST_MYSQL_HOST=localhost
```
nabdelgadir: Why are these lines commented?

.travis.yml
dhmlau: thx

agnes512:
@slnode test please

b-admike:
@slnode test please

agnes512:
@slnode test please

bajtos:
Can we use the MySQL server provided by Travis CI, instead of installing one via `apt-get`? See https://docs.travis-ci.com/user/database-setup/#mysql I would hope that the built-in MySQL instance will be faster too bootstrap.

Other than that, the changes LGTM.

b-admike:
> Can we use the MySQL server provided by Travis CI, instead of installing one via `apt-get`? See https://docs.travis-ci.com/user/database-setup/#mysql I would hope that the built-in MySQL instance will be faster too bootstrap.
> 
> Other than that, the changes LGTM.

Good point. We wanted to pin the MySQL version to `5.7` and thus the extra `addons` section as suggested in https://docs.travis-ci.com/user/database-setup/#mysql-57. But I also see that it's the default for the `Xenial` image, so we can just use that for the build instead 👍.

Issue #386, opened 2019-05-20
Run shared tests from both v3 and v4 of juggler


bajtos:
At the moment, our connectors are installing juggler 3.x to run the test suite. When we make a change to juggler@3, cis-jenkins detects downstream dependency and triggers CI build of connectors with the modified juggler version. When we make a change to juggler's master, no downstream builds are triggered.

In the past, we were maintaining multiple branches to test connectors against different juggler versions. E.g. loopback-2.x containing code from "master" but installing juggler 2.x for testing. This is rather impractical, because we don't have bandwidth to update those other branches with changes made to master.

Let's rework the connector test setup to execute tests from both juggler versions 3.x and 4.x, as we are already doing in other connector repositories.

The reference pull request to draw inspiration from: strongloop/loopback-connector-mongodb#519. See also strongloop/loopback-connector-postgresql#376.

Obviously, we need to fix the current test setup to pass on CI first, see https://github.com/strongloop/loopback-connector-mysql/issues/378

jannyHou:
PR merged

PR #390, opened 2019-06-28
feat: run shared tests

jannyHou:
### Description

connect to https://github.com/strongloop/loopback-connector-mysql/issues/386

Run shared tests from juggler v3 and v4.

#### Related issues
https://github.com/strongloop/loopback-connector-mysql/issues/386

### Checklist


package-lock.json
```diff
@@ -0,0 +1,1902 @@
+{
```
rmg: I strongly recommend _not_ committing a `package-lock.json` to this repo. It will lock the dependencies used by CI and loopback-connector-mysql devs while _users_ of loopback-connector-mysql will get newer/updated versions of the dependencies.

This means that CI will not be testing what users are using and may missing important bugs introduced by upstream dependencies.

jannyHou: True, deleted 🙂 it's auto generated by my PR, now tests run successfully, thanks!

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "6"
   - "8"
   - "10"
```
bajtos: Can you add Node.js 12 while changing `.travis.yml`? Not a big deal, we can add it later in a new PR.

lib/mysql.js
```diff
@@ -444,6 +444,12 @@ MySQL.prototype.fromColumnValue = function(prop, val) {
           lat: val.y,
         };
         break;
+      case 'ObjectID':
```
bajtos: Interesting. Why do we need to deal with `ObjectID`, considering that it's a type specific to MongoDB? Is this related to changes made in juggler v4?

lib/mysql.js
```diff
@@ -453,10 +459,6 @@ MySQL.prototype.fromColumnValue = function(prop, val) {
         }
         break;
       default:
-        if (!Array.isArray(prop.type) && !prop.type.modelName) {
-          // Do not convert array and model types
```
bajtos: If you remove this `default` case, does it mean that we are no longer going to convert other types? What are the implications for existing applications, can we break any edge cases?

test/init.js
```diff
@@ -6,8 +6,8 @@
 'use strict';
 
 module.exports = require('should');
-
-var DataSource = require('loopback-datasource-juggler').DataSource;
+const juggler = require('loopback-datasource-juggler');
+let DataSource = require('loopback-datasource-juggler').DataSource;
```
bajtos:
```suggestion
let DataSource = juggler.DataSource;
```

test/datatypes.test.js
```diff
@@ -206,7 +207,8 @@ describe('MySQL specific datatypes', function() {
     var xcor, ycor;
     City.create(city1, function(err, res) {
       if (err) return done(err);
-      res.loc.should.deepEqual(city1.loc);
+      const loc_in_geo_type = new GeoPoint(city1.loc);
+      res.loc.should.deepEqual(loc_in_geo_type);
```
bajtos: This change seems unrelated to the introduction of shared test suite from juggler v4. Is it needed to fix failing tests? Can we make it in a standalone PR please?

test/migration.test.js
```diff
@@ -639,7 +639,9 @@ var getFields = function(model, cb) {
       res.forEach(function(field) {
         fields[field.Field] = field;
       });
-      cb(err, fields);
+      // The returned data are in arrays of type `RowDataPacket`,
+      // which are not objects.
+      cb(err, JSON.parse(JSON.stringify(fields)));
```
bajtos: This change seems unrelated to the introduction of shared test suite from juggler v4. Is it needed to fix failing tests? Can we make it in a standalone PR please?

test/mysql.test.js
```diff
@@ -88,7 +89,7 @@ describe('mysql', function() {
 
         p.content.should.be.equal(post.content);
         p.title.should.be.equal('a');
-        p.comments.should.eql(['1', '2']);
+        p.comments.should.eql(new List(['1', '2']));
```
bajtos: This change seems unrelated to the introduction of shared test suite from juggler v4. Is it needed to fix failing tests? Can we make it in a standalone PR please?

jannyHou:
https://github.com/strongloop/loopback-connector-mysql/pull/390#discussion_r301130833
https://github.com/strongloop/loopback-connector-mysql/pull/390#discussion_r301130783
https://github.com/strongloop/loopback-connector-mysql/pull/390#discussion_r301130754

I submitted another PR to verify are they working with juggler v3 in

.travis.yml
jannyHou: @bajtos Yep I can submit another PR to add Node.js 12, I moved 6 in this PR since CI fails.

Running juggler v3 and v4 causes 20+ failures, I am not sure are they all broken by the upgrading of juggler, but this PR and https://github.com/strongloop/loopback-connector-mysql/pull/392 needs to fix them all, which is already a big amount of work, so I would like to add Node 12 in a separate PR :)

lib/mysql.js
jannyHou: @bajtos See tests in https://github.com/strongloop/loopback-connector-mysql/blob/1849507e065985ea0242440979a073bcebe4682e/test/mysql.test.js#L14, I guess it's for consistent behavior across all connectors.

lib/mysql.js
jannyHou: @bajtos The fix here is tricky, I have spent lots of time to figure out what is the expected behaviour across all the connectors:

When running tests in juggler v4, the type conversion here causes https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/datatype.test.js#L193-L199 fail. This test expects the connector returns the value from db AS IT IS.

And if you look at the postgresql connector https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/postgresql.js#L246, it also returns the value WITHOUT conversion.

Therefore I believe at least for juggler v4, the expected behaviour is connectors don't do type conversions for unknown types.

jannyHou:
@bajtos [Those 3 tests](https://github.com/strongloop/loopback-connector-mysql/pull/390#issuecomment-509319325) are related to the v3->v4 upgrade of juggler, see PR https://github.com/strongloop/loopback-connector-mysql/pull/392, so those fixes should still be kept in this PR.

I think v4 has a breaking change for the type coercion.

jannyHou:
@bajtos Thank you for the review, replied.

I am ok to release a new major version for upgrading to juggler v4. For the failures on the current master,  I prefer to leave it out of the scope of issue https://github.com/strongloop/loopback-connector-mysql/issues/386, we can either reopen https://github.com/strongloop/loopback-connector-mysql/issues/378 or create a new task for it :)

.travis.yml
bajtos: Makes sense :+1:

lib/mysql.js
bajtos: I see. `test/mysql.test.js` is not part of the shared test suite imported from juggler, so why are we fixing it in this pull request? Can you please open a new pull requests to fix problems that are unrelated to the shared test suite first?

lib/mysql.js
bajtos: I see, thank you for explaining the details to me.

bajtos:
I am not sure if we need a new major version, leave that up to you to decide.

I think I am starting to understand what's the problem here. We are not only adding the shared test suite from juggler v4, but also upgrading the mysql-connector-specific test suite to execute against juggler v4. On `master`, these connector-specific tests are executed against juggler v3.

In that case it makes sense to keep the changes in `test/*` files as part of this pull request 👍 
- https://github.com/strongloop/loopback-connector-mysql/pull/390#discussion_r301130833
- https://github.com/strongloop/loopback-connector-mysql/pull/390#discussion_r301130783 
- https://github.com/strongloop/loopback-connector-mysql/pull/390#discussion_r301130754

lib/mysql.js
jannyHou: @bajtos Interesting 🤔 I thought you have noticed that after configuring the shared tests, the connector tests run with juggler v4(which this PR does) instead of juggler v3(the current master), and that's how all these fixes happen. 
And I thought "running connector tests with juggler v4" is what we intend to do...

jannyHou:
@bajtos 
> I think I am starting to understand what's the problem here. We are not only adding the shared test suite from juggler v4, but also upgrading the mysql-connector-specific test suite to execute against juggler v4. On master, these connector-specific tests are executed against juggler v3.

Ah yes! Then never mind about my comment https://github.com/strongloop/loopback-connector-mysql/pull/390#discussion_r301623757.

PR #401, opened 2019-11-05
Fix tests

jannyHou:
### Description

Fixing 2 failures in https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-mysql~master/130/console
#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "juggler-v3": "file:./deps/juggler-v3",
     "juggler-v4": "file:./deps/juggler-v4",
     "loopback-datasource-juggler": "^3.0.0 || ^4.0.0",
-    "mocha": "^2.1.0",
+    "mocha": "^6.2.2",
     "rc": "^1.0.0",
     "should": "^8.0.2",
     "sinon": "^1.15.4"
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "pretest": "node pretest.js",
     "lint": "eslint .",
-    "test": "mocha --timeout 10000  test/*.test.js node_modules/juggler-v3/test.js node_modules/juggler-v4/test.js",
+    "test": "mocha --timeout 10000 --exit test/*.test.js node_modules/juggler-v3/test.js node_modules/juggler-v4/test.js",
     "posttest": "npm run lint"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


test/datetime.test.js
```diff
@@ -39,14 +39,15 @@ describe('MySQL datetime handling', function() {
     setConnectionTimezones('Z');
   });
   after(function(done) {
+    console.log('after');
```
raymondfeng: Please remove console.log.

test/transaction.promise.test.js
```diff
@@ -176,7 +176,8 @@ describe('transactions with promise', function() {
     it('should rollback the transaction if timeout', function(done) {
       Post.find({where: {title: 't3'}}, {transaction: currentTx},
         function(err, posts) {
-          if (err) return done(err);
+          should.exist(err);
```
raymondfeng: This change is suspicious. We expect to rollback an active transaction upon timeout.

jannyHou: @raymondfeng yeah I know:
My local tests with mysql 5.6.17 passes this test, **only CI fails**, that's why I make this change.

And according to [this line](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L153) and [the ensureActive method](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/transaction.js#L201), an error should be thrown here. 

Any suggestion?

jannyHou:
@slnode test please

test/transaction.promise.test.js
jannyHou: never mind I think it's related to the `--exit` flag :p

package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "pretest": "node pretest.js",
     "lint": "eslint .",
-    "test": "mocha --timeout 10000  test/*.test.js node_modules/juggler-v3/test.js node_modules/juggler-v4/test.js",
+    "test": "mocha --timeout 10000 --exit test/*.test.js node_modules/juggler-v3/test.js node_modules/juggler-v4/test.js",
```
bajtos: @jannyHou please move this flag to https://github.com/strongloop/loopback-connector-mysql/blob/master/test/mocha.opts

Issue #424, opened 2020-07-24
Bug to generate the Foreign key constraint during a migration if the property doesn't have the same name that the mysql colu


oom-:
##  Steps to reproduce

Create a model with foreign key + `belongsTo` decorator 
![image](https://user-images.githubusercontent.com/11071671/88406892-9632ec00-cdd1-11ea-9a2d-99cde775e1ef.png)

## Current Behavior

The migrate script crash because he looks for the name of the foreign key inside the props name of the object.
![image](https://user-images.githubusercontent.com/11071671/88407059-d4c8a680-cdd1-11ea-8b9a-555ed20b415b.png)

## Expected Behavior

Should just return the propName if not found instead of looking for `mysql` prop.

## Additional information

`win32 x64 12.16.1`
I will try to propose a fix

```js
 if (typeof modelToCheck.properties[propName] == 'undefined')
      return propName;
```

agnes512:
@oom- thank you for your investigation! I was able to locate the problem real quick. I have submitted a PR to address the issue, thanks!

agnes512:
@oom- I double checked it and found that there's something wrong with your relation definition. Since your foreign key is different from the default one, you will need to specify **the relation name** (see [docs](https://loopback.io/doc/en/lb4/BelongsTo-relation.html#relation-metadata)). For your case, it would be
```ts
@model({
  settings:{
        foreignKeys: {
      fk_order_customerId: {
        name: 'fk_branch_legalentity',
        entity: 'LegalEntity',
        entityKey: 'leId', // uses the PROP NAME instead of the actual column name
        foreignKey: 'leId',
      },
    },
  }
})
  @belongsTo(() => LegalEntity,
    {name: 'branch'}, // the relation name
    {mysql:{columnName: 'LE_ID'}}
    )
    leId: number;
```
Sorry that for some reasons I didn't spot it the first time 🤦  Could you try the definition and see if it works on your end? I am closing those PRs, and they can be reopened if it is actual a bug

mathias22osterhagen22:
Hello, it doesn't work:

![image](https://user-images.githubusercontent.com/60438474/90518955-b30bd680-e167-11ea-9077-16ba55fb3b85.png)

![image](https://user-images.githubusercontent.com/60438474/90518991-c159f280-e167-11ea-97ec-e2fe5eef9b9f.png)

agnes512:
@mathias22osterhagen22 is this the same error you got before?

[...]

PR #427, opened 2020-08-04
fix: allow to generate foreign key constraint with custom fk name

agnes512:
Fixes https://github.com/strongloop/loopback-connector-mysql/issues/424

Fixes the issue that the fk constraint cannot be generated if the fk name is customized.

Once it is fixed, we only need to specify the custom name in the **property definition** and use the property name as is in the foreign key constraint settings. E.g:
```ts
@model({
  settings:{
    'foreignKeys': {
      'fk_customer_customerId': {
        'name': 'fk_customert_customerId',
        'entity': 'Customer',
        'entityKey': 'id',
        'foreignKey': 'customerId', // property name
      },
  }
})
export class Order extends Entity {
  @property({
    type: 'number',
    id: true,
    generated: true,
  })
  id?: number;

  @belongsTo(() => Customer,
    {},
    {mysql:{columnName: 'MY_FK'}} // custom name
    )
  customerId: number;
```

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-connector-mysql) 👈

- [x] `npm test` passes on your machine
- [ ] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [ ] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -1022,14 +1022,16 @@ function mixinMigration(MySQL, mysql) {
     return columnType;
   }
   function expectedColNameForModel(propName, modelToCheck) {
-    var mysql = modelToCheck.properties[propName].mysql;
-    if (typeof mysql === 'undefined') {
-      return propName;
+    if (modelToCheck.properties[propName].mysql) {
+      var mysql = modelToCheck.properties[propName].mysql;
+      if (typeof mysql === 'undefined') {
+        return propName;
+      }
+      var colName = mysql.columnName;
+      if (typeof colName !== 'undefined') {
+        return colName;
+      }
     }
-    var colName = mysql.columnName;
-    if (typeof colName === 'undefined') {
-      return propName;
-    }
-    return colName;
+    return propName;
   }
 }
--- a/test/mysql.autoupdate.test.js
+++ b/test/mysql.autoupdate.test.js
@@ -788,6 +788,93 @@ describe('MySQL connector', function() {
     });
   });
 
+  it('should auto migrate/update custom foreign keys name with its constraint', function(done) {
+    var sourceModel = {
+      'name': 'SourceModel',
+      'options': {
+        'idInjection': false,
+        'mysql': {
+          'schema': 'myapp_test',
+          'table': 'source_model',
+        },
+      },
+      'properties': {
+        'id': {
+          'type': 'String',
+          'length': 20,
+          'id': 1,
+        },
+        'name': {
+          'type': 'String',
+          'length': 20,
+        },
+      },
+    };
+
+    var customFkModel = {
+      'name': 'CustomFkTest',
+      'options': {
+        'idInjection': false,
+        'mysql': {
+          'schema': 'myapp_test',
+          'table': 'custom_fk_test',
+        },
+        'foreignKeys': {
+          'fk_customfktest_customerId': {
+            'name': 'fk_customfktest_customerId',
+            'entity': 'SourceModel',
+            'entityKey': 'id',
+            'foreignKey': 'customerId',
+          },
+        },
+      },
+      'properties': {
+        'id': {
+          'type': 'String',
+          'length': 20,
+          'id': 1,
+        },
+        'customerId': {
+          'type': 'String',
+          'length': 20,
+          'mysql': {'columnName': 'custom_fk'},
+        },
+      },
+    };
+
+    var foreignKeySelect =
+    'SELECT COLUMN_NAME,CONSTRAINT_NAME,REFERENCED_TABLE_NAME, REFERENCED_COLUMN_NAME ' +
+    'FROM   INFORMATION_SCHEMA.KEY_COLUMN_USAGE ' +
+    'WHERE  REFERENCED_TABLE_SCHEMA = "myapp_test" ' +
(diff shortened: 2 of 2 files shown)


agnes512:
@slnode test please

agnes512:
@slnode test please

lib/migration.js
```diff
@@ -1022,14 +1022,16 @@ function mixinMigration(MySQL, mysql) {
...
-    var mysql = modelToCheck.properties[propName].mysql;
-    if (typeof mysql === 'undefined') {
-      return propName;
-    }
-    var colName = mysql.columnName;
-    if (typeof colName === 'undefined') {
-      return propName;
+    if (modelToCheck.properties[propName].mysql) {
+      var mysql = modelToCheck.properties[propName].mysql;
+      if (typeof mysql === 'undefined') {
```
jannyHou, line 1027: I feel the logic here could be simplified:
`if(modelToCheck.properties[propName].mysql)` means `if (typeof mysql === 'undefined')` won't happen.

Or did I miss something?

jannyHou: I am a bit confused what changes here:

Before: 
- If `modelToCheck.properties[propName].mysql` exists returns propName
- If `mysql.columnName` exists returns colName

And the new code also does the same thing...?

test/mysql.autoupdate.test.js
```diff
@@ -788,6 +788,93 @@ describe('MySQL connector', function() {
...
+    'WHERE  REFERENCED_TABLE_SCHEMA = "myapp_test" ' +
+    'AND   TABLE_NAME = "custom_fk_test"';
+    var getCreateTable = 'SHOW CREATE TABLE `myapp_test`.`custom_fk_test`';
+
+    ds.createModel(sourceModel.name, sourceModel.properties, sourceModel.options);
+    ds.createModel(customFkModel.name, customFkModel.properties, customFkModel.options);
+
+    // do initial update/creation of table
+    ds.autoupdate(function(err) {
+      assert(!err, err);
```
bajtos, line 857: Errors should be propagated to the test runner, so that the test failure includes the original error stack trace. When using `assert`, the original stack trace is discarded and the test failure shows a stack pointing to the assertion line.

```suggestion
      if (err) return done(err);
```

Please note that we usually use async/await style in new tests.

```js
it('should auto migrate/update custom foreign keys name with its constraint', async () => {
  // ...
  const props = await ds.discoverModelProperties('custom_fk_test');
  // ...
});
```

test/mysql.autoupdate.test.js
```diff
@@ -788,6 +788,93 @@ describe('MySQL connector', function() {
...
+      ds.discoverModelProperties('custom_fk_test', function(err, props) {
+        // validate that we have the correct number of properties
+        assert.equal(props.length, 2);
```
bajtos, line 860: It's better to use `should` than `assert` as it produces more helpful error messages on failure.

```suggestion
        props.should.have.length(2);
```

The same comment applies to other assertions below too.

test/mysql.autoupdate.test.js
```diff
@@ -788,6 +788,93 @@ describe('MySQL connector', function() {
...
+
+        // get the foreign keys for this table
+        ds.connector.execute(foreignKeySelect, function(err, foreignKeys) {
+          if (err) return done(err);
+          // validate that the foreign key exists and points to the right column
+          assert(foreignKeys);
+          assert(foreignKeys.length.should.be.equal(1));
```
bajtos, line 867:
```suggestion
          foreignKeys.should.have.length(1);
```

test/mysql.autoupdate.test.js
```diff
@@ -788,6 +788,93 @@ describe('MySQL connector', function() {
...
+          assert.equal(foreignKeys[0].REFERENCED_TABLE_NAME, 'source_model');
+          assert.equal(foreignKeys[0].COLUMN_NAME, 'custom_fk');
+          assert.equal(foreignKeys[0].CONSTRAINT_NAME, 'fk_customfktest_customerId');
+          assert.equal(foreignKeys[0].REFERENCED_COLUMN_NAME, 'id');
```
bajtos, line 871:
```suggestion
          foreignKeys[0].should.containEql({
            REFERENCED_TABLE_NAME: 'source_model',
            COLUMN_NAME: 'custom_fk',
            CONSTRAINT_NAME: 'fk_customfktest_customerId',
            REFERENCED_COLUMN_NAME: 'id',
          });
```

lib/migration.js
bajtos, line 1027: Can we keep the original style please? Also we use `const` instead of var in new code.

```suggestion
    const mysql = modelToCheck.properties[propName].mysql;
    if (typeof mysql === 'undefined') {
```

agnes512: You're right.. I just tried and couldn't to reproduce the issue 🙈 ... I think the original issue sets the relation incorrectly.

agnes512:
Sorry, false alarm, see https://github.com/strongloop/loopback-connector-mysql/issues/424#issuecomment-673180986

PR #594, opened 2024-06-25
fix: mark primary key auto-increment only if column type is INT

aaqilniz:
This PR marks an id property autoincrement only when type of the id property is integer/number.

## Checklist

- [x] DCO (Developer Certificate of Origin) [signed in all commits](https://loopback.io/doc/en/contrib/code-contrib.html)
- [x] `npm test` passes on your machine
- [x] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [x] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -628,7 +628,13 @@ function mixinMigration(MySQL, mysql) {
       if (pks.length === 1) {
         const idName = this.idName(model);
         const idProp = this.getModelDefinition(model).properties[idName];
-        if (idProp.generated) {
+        const idColumnType = this.columnDataType(model, idName);
+        if (idProp.generated && (
+          idColumnType === 'TINYINT' ||
+          idColumnType === 'SMALLINT' ||
+          idColumnType === 'MEDIUMINT' ||
+          idColumnType === 'INT' ||
+          idColumnType === 'BIGINT')) {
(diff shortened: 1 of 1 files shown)


lib/migration.js
```diff
@@ -628,7 +628,8 @@ function mixinMigration(MySQL, mysql) {
       if (pks.length === 1) {
         const idName = this.idName(model);
         const idProp = this.getModelDefinition(model).properties[idName];
-        if (idProp.generated) {
+        const idColumnType = this.columnDataType(model, idName);
+        if (idProp.generated && idColumnType === 'INT') {
```
dhmlau, line 632: I believe there are other integer type that supports auto increment in mysql, according to this docs: https://dev.mysql.com/doc/refman/8.4/en/example-auto-increment.html#:~:text=Use%20the%20smallest%20integer%20data,to%20allow%20a%20greater%20range.

aaqilniz: You are right @dhmlau. The other integer types also support autoincrement. I have updated the PR. Please have a look at the PR now.

Thanks for the review.

dhmlau (approved):

PR #608, opened 2024-09-01
fix: mark a property generated if it is STORED or VIRTUAL

aaqilniz:
## Checklist

- [x] DCO (Developer Certificate of Origin) [signed in all commits](https://loopback.io/doc/en/contrib/code-contrib.html)
- [ ] `npm test` passes on your machine
- [x] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [x] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)


lib/discovery.js
```diff
@@ -181,7 +188,14 @@ function mixinDiscovery(MySQL, mysql) {
         ' numeric_scale AS "dataScale",' +
         ' column_type AS "columnType",' +
         ' is_nullable = \'YES\' AS "nullable",' +
-        ' CASE WHEN extra LIKE \'%auto_increment%\' THEN 1 ELSE 0 END AS "generated"' +
+        `
+          case
```
dhmlau, line 192: If there's `GENERATED ALWAYS` when using `VIRTUAL` or `STORED`, should we check for that? 

Also could you please make the syntax consistent? i.e. for line 193:
```
 WHEN extra LIKE '%virtual%' THEN 1
```

dhmlau (commented):
@aaqilniz, I'm not familiar with mysql, hope my comment make sense. Thanks.

lib/discovery.js
dhmlau: @aaqilniz, are you interested to continue to pursue with this PR?

aaqilniz: Hi, @dhmlau. I will update the PR soon and notify you for the review.

aaqilniz: Hi, @dhmlau. I have just made a quick update. Can you please have a look at the PR?

